PR #9, opened 2019-07-22
SDN role/binding cleanup

danwinship:
These roles have been unused since 3.11, because CNO creates its own roles for openshift-sdn to use (if and only if openshift-sdn is being used).

Formerly https://github.com/openshift/origin/pull/23184 but openshift-apiserver appears to be vendored now (contrary to the repo description at the top of https://github.com/openshift/openshift-apiserver).


sttts:
Note that these roles don't go away automatically in existing clusters. If you want that as well, add deletion code to your operator.

sttts:
`system:sdn-reader` and `system:sdn-manager` are not only meant to be used by our operators. Customers might depend on them.

Are these documented anyway? Do we accept that the customer cannot use them anymore?

At the very least we should have release notes about this.

danwinship:
> `system:sdn-reader` and `system:sdn-manager` are not only meant to be used by our operators. 

I don't think that's true. We never considered what set of permissions might be needed by arbitrary SDNs; they were very specifically the exact set of permissions that were needed by openshift-sdn and nothing else. (OVN-Kubernetes, for example, can not run with those permissions.) I'm pretty sure we removed some excess permissions from them at one point to tighten up security.

> Are these documented anyway?

No.

> Note that these roles don't go away automatically in existing clusters. If you want that as well, add deletion code to your operator.

My operator didn't create them, yours did. :stuck_out_tongue_closed_eyes: 

Anyway, TL;DR: no part of OpenShift 4.x has ever used these roles. If you want to keep creating them, go ahead, I guess.

sttts:
> I don't think that's true. We never considered what set of permissions might be needed by arbitrary SDNs; they were very specifically the exact set of permissions that were needed by openshift-sdn and nothing else. (OVN-Kubernetes, for example, can not run with those permissions.) I'm pretty sure we removed some excess permissions from them at one point to tighten up security.

I was meaning RBAC roles in general with "not only meant to be used by operator". They are public and available to customers. There might be customers out there who depend on them. I think a release note is needed at the very least.

> My operator didn't create them, yours did. 😝

Not even the operator, the apiserver, which was originally in origin, shared repo, shared responsibility. We don't own all the APIs in openshift-apiserver, only the mechanical bits. The teams owning the APIs also own the RBAC rules. If you care, PR welcome. Otherwise, old clusters have some garbage rules.

danwinship:
OK, I've figured out where the 4.4 release notes are, so if you merge this, I'll add a note there.

PR #11, opened 2019-07-29
switch to apiserver-library-go

deads2k:
Checking what is necessary to drop our kube fork here.


deads2k:
@sttts got it.  I think this is the way to go.

PR #14, opened 2019-08-07
[PROOF] UPSTREAM: 00000: add dynamic certificate reloading

mfojtik:
This was missed during split and I have theory that is what cause https://bugzilla.redhat.com/show_bug.cgi?id=1736800

It looks like that after rotation, the dynamic reload in openshift apiserver won't pick up the new client certificates and wedge until somebody kick those pods and then restart.

/cc @deads2k 
/cc @sttts 

@deads2k I don't know how we are going to pick this, but I assume https://github.com/openshift/kubernetes-apiserver/pull/1 is the way?

@p0lyn0mial @stlaz can one of you check @deads2k patches and double-check that we indeed missing this?


p0lyn0mial:
I think that this commit is exactly the same as the one committed to `origin` and indeed it is missing.

mfojtik:
/hold

This is just a proof that the carry patch won't break anything.

stlaz:
@mfojtik this PR only seems to add dynamic reload of serving certs, which we do need, but to fix the trouble connecting to the Kube APIserver, we also need to include https://github.com/openshift/origin/pull/22229/ which appears to be reloading the client-CA, too

vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go
```diff
@@ -67,16 +66,16 @@ func (s *SecureServingInfo) Serve(handler http.Handler, shutdownTimeout time.Dur
...
-	// cert will become the default cert. That's what we expect anyway.
-	for _, c := range s.SNICerts {
-		secureServer.TLSConfig.Certificates = append(secureServer.TLSConfig.Certificates, *c)
+	// this option overrides the provided certs
+	// TODO this should be mutually exclusive, but I'm not sure what that will do today
+	if s.DynamicCertificates != nil {
+		secureServer.TLSConfig.Certificates = nil
+		// need to load the certs at least once
+		if err := s.DynamicCertificates.CheckCerts(); err != nil {
+			return err
```
stlaz: We're missing some more patches, the build failure in https://github.com/openshift/openshift-apiserver/pull/15 proves that the bits in this PR were not used since we're not actually using the repositories in the same manner as in `origin` (no symlinks to kube staging repos).
Notice the `return err` here when the function type signature demands `(<-chan struct{}, error)`

mfojtik:
https://github.com/openshift/openshift-apiserver/pull/16

PR #15, opened 2019-08-08
[PROOF2] UPSTREAM: 00000: add dynamic certificate reloading 

stlaz:
This PR merely takes https://github.com/openshift/openshift-apiserver/pull/14 and adds it to the appropriate location, I don't think vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver will ever be used (since we don't symlink to it)

Let's see if it breaks stuff. I'll test that locally with my reproducer, too. I think we still might need to add the client CA bits from origin but I'll test the minimal version first.


stlaz:
Fair amount of manual work was done to get those patches in, and that's just to make this build. I'm a bit scared about all the things that we might be missing in the openshift-apiserver.

stlaz:
OpenShift apiserver pods fail with x509 errors even after this patchset, I'll look into additional ones.

stlaz:
superseeded by https://github.com/openshift/openshift-apiserver/pull/16

PR #23, opened 2019-09-17
Bug 1741391: support ImageContentSourcePolicy by ImageStreamImport

dmage:


dmage:
@sttts this PR is expected eventually to be backported to 4.2.z. Is it ok to change vendor for such PRs?

pkg/image/apiserver/importer/importer.go
```diff
@@ -389,6 +392,121 @@ func (isi *ImageStreamImporter) calculateImageSize(ctx gocontext.Context, bs dis
...
+				bestMatch = &reg
+			}
+		}
+	}
+	return bestMatch
+}
+
+func (isi *ImageStreamImporter) getManifestByDigestFromMirror(ctx gocontext.Context, retriever RepositoryRetriever, ref imageapi.DockerImageReference, d godigest.Digest, insecure bool) (distribution.Manifest, distribution.ManifestService, distribution.BlobStore) {
+	reg := isi.findRegistryConfiguration(ref)
+	if reg == nil {
```
mtrmac: I don’t think `getManifestByDigestFromMirror` and `getManifestByDigestFromSource` need to be separate; instead, just build a single-entry `pullSources` here, and handle it through the loop below.

pkg/image/apiserver/importer/importer.go
```diff
@@ -389,6 +392,121 @@ func (isi *ImageStreamImporter) calculateImageSize(ctx gocontext.Context, bs dis
...
+	}
+
+	pullSources, err := reg.PullSourcesFromReference(dockerRef)
+	if err != nil {
+		klog.V(5).Infof("unable to get alternative pull sources for %s: %v", ref.Exact(), err)
+		return nil, nil, nil
+	}
+
+	for _, pullSource := range pullSources {
+		mirrorRef, err := imageref.Parse(pullSource.Reference.String())
```
mtrmac: Throughout this loop: the last entry of `pullSources` is the upstream registry, so all of the references to “mirrors” in variable names and error messages might be confusing.

(Also, no need to call `getManifestByDigestFromSource` if this loop fails for all items.)

pkg/image/apiserver/importer/importer.go
```diff
@@ -462,50 +580,62 @@ func (isi *ImageStreamImporter) importManifest(ctx gocontext.Context, manifest d
 // optional rate limiter.  Errors are set onto the individual tags and digest objects.
 func (isi *ImageStreamImporter) importRepositoryFromDocker(ctx gocontext.Context, retriever RepositoryRetriever, repository *importRepository, limiter flowcontrol.RateLimiter) {
 	klog.V(5).Infof("importing remote Docker repository registry=%s repository=%s insecure=%t", repository.Registry, repository.Name, repository.Insecure)
-	// retrieve the repository
-	repo, err := retriever.Repository(ctx, repository.Registry, repository.Name, repository.Insecure)
-	if err != nil {
-		klog.V(5).Infof("unable to access repository %#v: %#v", repository, err)
-		switch {
-		case err == reference.ErrReferenceInvalidFormat:
```
mtrmac: This detailed error handling seems to have been lost in the move to `getManifestByDigest`.

pkg/image/apiserver/importer/importer.go
```diff
@@ -462,50 +580,62 @@ func (isi *ImageStreamImporter) importManifest(ctx gocontext.Context, manifest d
...
 		}
-		applyErrorToRepository(repository, err)
-		return
-	}
 
-	// get a manifest context
-	s, err := repo.Manifests(ctx)
-	if err != nil {
-		klog.V(5).Infof("unable to access manifests for repository %#v: %#v", repository, err)
-		switch {
```
mtrmac: This detailed error handling seems to have been lost in the move to `getManifestByDigest`.

pkg/image/apiserver/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,260 @@
+package sysregistriesv2
```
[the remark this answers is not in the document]
dmage: Because the original package also has V1 types that I don't need. And these types need additional dependencies that may create unnecessary complications for backporting this PR (types.SystemContext, for example).

@adambkaplan It'd be good to make packages more granular in the new repo, so we don't need to import dozen packages only because we need to merge two maps.

pkg/image/apiserver/registries/registries.go
```diff
@@ -0,0 +1,164 @@
...
+	return res, nil
+}
+
+// EditRegistriesConfig edits, IN PLACE, the /etc/containers/registries.conf configuration provided in config, to:
+// - Mark whole registries in insecureRegistries as insecure (TLS is not required, and TLS certificate verification is not required when TLS is used)
+// - Mark whole registries in blockedRegistries as blocked (any attempts to access them fail)
+// - Implement ImageContentSourcePolicy rules in icspRules.
+// “Whole registries” above means that the configuration applies to everything on that registry, including any possible separately-configured
+// namespaces/repositories within that registry.
+func EditRegistriesConfig(config *sysregistriesv2.V2RegistriesConf, insecureRegistries, blockedRegistries []string, icspRules []*apioperatorsv1alpha1.ImageContentSourcePolicy) error {
```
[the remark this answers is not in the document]
dmage: I don't want to vendor any operator into the API server. The repo is not ready yet. But even if this code live in a new repo, I still don't want to vendor it because of its dependency on the all-in-one package sysregistriesv2. It will add dependency on toml, for example, that is not needed for the API server.

/cc @adambkaplan

pkg/image/apiserver/importer/importer.go
```diff
@@ -389,6 +392,121 @@ func (isi *ImageStreamImporter) calculateImageSize(ctx gocontext.Context, bs dis
...
+	repoName := ref.AsRepository().Exact()
+
+	var bestMatch *sysregistriesv2.Registry
+	for i := range isi.regConf.Registries {
+		// XXX We cannot use `i, reg := range` here and `&reg` later.
+		// Otherwise the value will be changed on the next loop iteration.
+		reg := isi.regConf.Registries[i]
+
+		if bestMatch == nil || len(reg.Prefix) > len(bestMatch.Prefix) {
+			if scopeMatchesRegistry(repoName, reg.Prefix) {
```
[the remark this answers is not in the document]
dmage: `repoName` doesn't contain `:` nor `@` here.

pkg/image/apiserver/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
mtrmac: Fair enough, https://github.com/containers/image/issues/692 really is a pain.

pkg/image/apiserver/importer/importer.go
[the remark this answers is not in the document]
mtrmac: Ah, true. The code is still misusing `scopeMatchesRegistry`, though.

pkg/image/apiserver/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
mtrmac: FWIW, after https://github.com/containers/image/pull/716 , c/image v4.0.0 no longer imports the compression implementations. Sure, there is still the TOML decoder, and of course `containers/image/types`, but I can’t see how merely including that code hurts all that much.

pkg/image/apiserver/registries/registries.go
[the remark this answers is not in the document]
adambkaplan: @mtrmac I'm getting this repo set up. We'll address the refactor in a later PR.

dmage:
@adambkaplan @sttts please take a look

pkg/image/apiserver/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
mtrmac: containers/image does not currently maintain per-OpenShift branches; you can vendor commit df216d7cdc943e3e2ddef7be67a6abdcaf583e61 directly.

(Or use use the later released v4.0.1, but that forces use of Go modules.)

mfojtik:
/hold

This is pending architectural discussion.

/cc @derekwaynecarr 
/cc @smarterclayton 
/cc @deads2k

smarterclayton:
So we want end users to be able to create ICSP.  It's dynamic global config that someone should be able to change without causing control plane restarts.

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -27,6 +28,7 @@ import (
 	"github.com/openshift/api/image"
 	imagev1 "github.com/openshift/api/image/v1"
 	imageclientv1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
+	operatorv1alpha1lister "github.com/openshift/client-go/operator/listers/operator/v1alpha1"
```
soltysh: v1 always

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -195,6 +202,22 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
+	v2regConf := &sysregistriesv2.V2RegistriesConf{}
+	err = registries.EditRegistriesConfig(v2regConf, nil, nil, icspRules)
+	if err != nil {
+		klog.Warningf("failed to merge ImageContentSourcePolicy resources, mirrored images will not be found: %v", err)
+	}
+	for i, reg := range v2regConf.Registries {
+		v2regConf.Registries[i].Prefix = reg.Location
+	}
+
+	klog.Infof("FINDME %#+v", v2regConf)
```
soltysh: Remove

pkg/image/apiserver/registry/imagestreamimport/rest.go
dmage: @soltysh for me it's a bit weird to name v1alpha1 as v1, but updated.

pkg/image/apiserver/registry/imagestreamimport/rest.go
dmage: Thanks, removed.

adambkaplan:
@mfojtik openshift-controller-manager does something similar with build controller configurations (proxy and runtime config in particular). In that case we had to tackle a few subtleties around data races that @deads2k pointed out. Here there is less of a concern since the ICSP is obtained from the lister cache with each imagestream import request.

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
```diff
@@ -9,6 +9,8 @@ import (
 	imagev1informer "github.com/openshift/client-go/image/informers/externalversions"
 	oauthv1client "github.com/openshift/client-go/oauth/clientset/versioned"
 	oauthv1informer "github.com/openshift/client-go/oauth/informers/externalversions"
+	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
```
adambkaplan: @dmage looks like this import is failing after your PR fast forwards on top of master. You may need to rebase to fix this.

adambkaplan:
@mfojtik per the email thread from @bparees, are we ok lifting the hold for this?

Summarizing the thread here - there is a broader TODO to define guardrails as to when it's appropriate for operands to consume cluster config vs. have an operator consume the configuration and roll out changes to an operand.

@dmage before giving the lgtm we need to ensure that we can debug this in a support/failure case:

1. Can a bad ICSP cause the apiserver to crashloop?
2. Can we sufficiently debug if a user reports that an imagestream is not importing from a mirror? (via must-gather data)

dmage:
@adambkaplan 

1. a bad ICSP shouldn't cause the apiserver to crashloop
2. I added some debug messages that can help us understand where the image is going to be pulled from

dmage:
Regarding debuggability:
1. now we have the image reference in all error message (https://github.com/openshift/openshift-apiserver/pull/23/files#diff-37ea299bfbc0e50a72dc84712ffb8076R372)
2. aggregate errors from all mirrors (https://github.com/openshift/openshift-apiserver/pull/23/files#diff-37ea299bfbc0e50a72dc84712ffb8076R543)
3. log references that we are trying to access (https://github.com/openshift/openshift-apiserver/pull/23/files#diff-37ea299bfbc0e50a72dc84712ffb8076R522)

dmage:
@sttts @mfojtik can we merge it or is there something left to fix?

mfojtik:
@dmage please rename the last commit to `bump(*): vendor`

adambkaplan:
/lgtm

Last commit renamed per @mfojtik

PR #24, opened 2019-09-17
[4.3] fallback to x86 arch for import before going to first image in the manifestlist

bparees:


bparees:
/hold

should not go in until 4.3 at the earliest.  also needs signoff from @smarterclayton that this is an acceptable api behavior change.

cc @adambkaplan @dmage

pkg/image/apiserver/importer/importer.go
```diff
@@ -408,6 +408,18 @@ func manifestFromManifestList(ctx gocontext.Context, manifestList *manifestlist.
 			break
 		}
 	}
+
+	// if we couldn't match the preferred arch/os, and we couldn't match the platform's arch/os, prefer
+	// x86/linux before falling back to "first image in the manifestlist" as a last resort.
+	if manifestDigest == "" {
+		for _, manifestDescriptor := range manifestList.Manifests {
+			if manifestDescriptor.Platform.Architecture == "amd64" && manifestDescriptor.Platform.OS == "linux" {
```
adambkaplan: @bparees where are we matching the "platform's arch/os"? Should we try and obtain this information from the cluster via a declared platform architecture?

bparees: immediately above this block.  

```
if preferArch == "" {
		preferArch = runtime.GOARCH
	}
	if preferOS == "" {
		preferOS = runtime.GOOS
	}
	var manifestDigest godigest.Digest
	for _, manifestDescriptor := range manifestList.Manifests {
		if manifestDescriptor.Platform.Architecture == preferArch && manifestDescriptor.Platform.OS == preferOS {
			manifestDigest = manifestDescriptor.Digest
			break
		}
	}
```

bparees: runtime.GOARCH/GOOS should reflect the architecture/os we are running on.

bparees:
i think i got an ok from @smarterclayton in slack, so
/assign @adambkaplan @dmage 
again, needs to hold until we branch for 4.3.

adambkaplan:
/lgtm

Keeping hold until 4.3

bparees:
@adambkaplan @dmage ready for this to be unheld?

PR #25, opened 2019-09-17
Bug 1752979: Retry imagestream imports

adambkaplan:
Retry an imagestream import if there is a network timeout.
Only retry once, and immediately if there is an import error.


adambkaplan:
/close

@dmage I didn't realize we are already retrying at the client level for network timeouts. If that's the case then there is likely a broader issue here (reliability of GCP's egress, registry.redhat.io., or both), and fixing in the imagestream import isn't the way to go.

PR #28, opened 2019-09-25
Bug 1755115: allow scheduled imagestream imports for dockerimages with sha/digest …

gabemontero:
…refs

per instructions from @smarterclayton  ... the api needs to be  “level driven”

with this change, we can use scheduled import to enact retry for imagestreams with dockerimage refs that container the canonical sha/digest

like the payload imagestreams oc, cli, must-gather, that reside in the openshift namespace

/assign @smarterclayton 
/assign @bparees 
/assign @dmage

@openshift/openshift-team-master FYI

### DIFF
--- a/pkg/image/apis/image/validation/validation.go
+++ b/pkg/image/apis/image/validation/validation.go
@@ -264,8 +264,6 @@ func ValidateImageStreamTagReference(
 		case "DockerImage":
 			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
 				errs = append(errs, field.Invalid(fldPath.Child("from", "name"), tagRef.From.Name, err.Error()))
-			} else if len(ref.ID) > 0 && tagRef.ImportPolicy.Scheduled {
-				errs = append(errs, field.Invalid(fldPath.Child("from", "name"), tagRef.From.Name, "only tags can be scheduled for import"))
 			} else if whitelister != nil {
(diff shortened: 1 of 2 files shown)


pkg/image/apis/image/validation/validation.go
```diff
@@ -264,8 +264,6 @@ func ValidateImageStreamTagReference(
 		case "DockerImage":
 			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
 				errs = append(errs, field.Invalid(fldPath.Child("from", "name"), tagRef.From.Name, err.Error()))
-			} else if len(ref.ID) > 0 && tagRef.ImportPolicy.Scheduled {
```
tnozicka: does the IS have to be scheduled to retry failed import? I'd assume the controller retries failures without a schedule

dmage: Yes, they do. If the requested image doesn't exist on the remote registry, generally speaking there is no reason to do another attempt. But the user may explicitly request to retry.

bparees: you'd be wrong :(

the image import api is a one-shot thing, if it fails, you get an imagestream w/ a failed import status and we never try again (unless scheduled import is enabled for the tag)

(we are discussing how that should change in the future but that's a longer term effort)

tnozicka: ok, thanks for the context

PR #29, opened 2019-09-27
Bug 1755115: allow scheduled imagestream imports for dockerimages with sha/digest …

gabemontero:
…refs

/assign @bparees 
@smarterclayton 
@dmage 

one more spot ... see https://github.com/openshift/cluster-samples-operator/pull/186#issuecomment-536036652

### DIFF
--- a/pkg/image/apis/image/validation/validation.go
+++ b/pkg/image/apis/image/validation/validation.go
@@ -537,12 +537,8 @@ func ValidateImageStreamImport(isi *imageapi.ImageStreamImport) field.ErrorList 
 				if spec.From.Name == "*" {
 					continue
 				}
-				if ref, err := imageref.Parse(spec.From.Name); err != nil {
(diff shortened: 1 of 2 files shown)


bparees:
seems like we're lacking sufficient test coverage of the validation behavior that we didn't catch this in your original PRs.

also can this code be combined w/ the other validation locations?

gabemontero:
> seems like we're lacking sufficient test coverage of the validation behavior that we didn't catch this in your original PRs.

Yep I'll see about adding a unit test

> 
> also can this code be combined w/ the other validation locations?

with the code removals that have occurred, there is only one spot now where we deal with the `Scheduled` field, and it is in `ValidateImageStreamTagReference`

`ValidateImageStreamImport` short circuits at a higher level and does not need to interrogate the `Scheduled` field

gabemontero:
test case added @bparees

PR #30, opened 2019-10-11
[proof] TLS 1.3 does not brake openshift-apiserver

stlaz:
/hold

Fake bumps library-go to see if adding 1.3 ciphers breaks the cluster or not. The ciphers should automatically be picked for the openshift-apiserver by `SetRecommendedGenericAPIServerConfigDefaults()`

openshift/library-go#553


stlaz:
/close
proven

PR #35, opened 2019-10-29
storage: setup encryption transformers if --encryption-provider-config is given

sttts:


sttts:
going with @p0lyn0mial's counterpart PR.

PR #47, opened 2019-11-20
remove side-channel wiring for cluster authentication

deads2k:
Eventually this should wire the standard option flags through, but for now simply removing the side-channel option is enough.


csrwng:
@deads2k @mfojtik @sttts This PR breaks the hosted control plane use case, since the openshift-apiserver no longer honors the authentication kubeconfig specified in the config.yaml. Specifying a KUBECONFIG env var does not fix it either. In a hosted control plane, the kube apiserver that openshift apiserver communicates with cannot be the inline one.

PR #48, opened 2019-11-21
Bug 1757553: bump apiserver-lib-go for escalating externalIPs

stlaz:


stlaz:
/hold
not seeing the change in vendor

stlaz:
/hold cancel
should be fine now

PR #51, opened 2019-11-24
Add the imagetags resource to v1

smarterclayton:
`imagetags` is a more general purpose version of `imagestreamtags` that is not impacted by the latter's complicated GET semantics (tags that have not yet been imported are not returned). Instead, imagetags is a slice of the image stream by a single tag, returning metadata from the stream, the spec and status fields (if set), and the last tagged image (if imported). Create on the image tag requires a valid spec field. Update of the image tag can clear or set the spec. The metadata of the image stream cannot be changed on update although a future update may allow that. Status may also not be changed on update, BUT the Update operation requires that status match (so that in the future we could allow some tags to be removed). Delete removes both spec and status.

```
kind: ImageTag
version: image.openshift.io/v1
metadata:
  name: a:b
  ... # same as image stream
spec:
  # optional, present if spec.tags[foo] exists, same value
status:
  # optional, present if status.tags[foo] exists, same value
image:
  # optional, equal to last resolved image on the tag, dockerImageReference set based on pull policy
```

`imagestreamtags` remains as is, but clients will be encouraged to use the simpler `imagetags` in the future.

All users with permission on `imagestreamtags` should be granted the same rights on `imagetags`.

The server side printers for imagetags provides much more context on the difference between an imported / import failed / referenced by digest / referenced by tag image, as well as showing count of history items and last update


smarterclayton:
```
$ oc get imagetag -n openshift
NAME                                               SPEC        STATUS                                                                          HISTORY   UPDATED
apicast-gateway:latest                             Push        image/sha256:460179d4b253aa5ec8230e1822b9a1d391f572b69b8339c8610598fe9e4b6e69   1         5 minutes ago
apicast-gateway:2.1.0.GA                           Tag         image/sha256:313df5722ddd866d43758af161e5932dfd4648b99a5c57acfddfc2a955669fe8   1         5 minutes ago
apicast-gateway:2.2.0.GA                           Tag         image/sha256:0a74fc0c17db354a09fba80d7733b0aad6a14ec665bcebcc43a00c6cdc0a3a48   1         5 minutes ago
apicast-gateway:2.3.0.GA                           Tag         image/sha256:1acc349748d65ca7bdd7689a2b5762e1bb81dd1ba7038e2a75c0f31099f3854d   1         5 minutes ago
apicast-gateway:2.4.0.GA                           Tag         image/sha256:76bebff9f4fb5c360974950e4977a710d18ca3ecee6a0dad9abf507ccd6f5f34   1         5 minutes ago
apicast-gateway:2.5.0.GA                           Tag         image/sha256:65ba79bf6f3f103f606011a63a12f48b8bbc4f49fde3f96bee55e975d35b00ff   1         5 minutes ago
apicast-gateway:2.6.0.GA                           Tag         image/sha256:460179d4b253aa5ec8230e1822b9a1d391f572b69b8339c8610598fe9e4b6e69   1         5 minutes ago
apicurito-ui:1.2                                   Tag         image/sha256:3e14960d3381cc5a9fd5072c07ea3147f4af33d96f38d8d3bada819c1112149d   1         4 minutes ago
apicurito-ui:1.3                                   Tag         image/sha256:9f86ab0b7ed4cb978398b1db47527ac4beefc796ce3d2476bef3d8573a82c641   1         4 minutes ago
apicurito-ui:1.4                                   Tag         image/sha256:cdcd8ceeadf7a6af6a3bf43f286b1219e9870595724f3235426fffb20f2a5b78   1         4 minutes ago
cli:latest                                         Scheduled   image/sha256:b1aa53d39a71b9eda0c72a837daabf489a833d157057d3f1d5ecbaceaab32cf7   1         6 minutes ago
cli-artifacts:latest                               Scheduled   image/sha256:f3dd12cb8df8806787ae2a320d80369f5f5d002a60029a2102809b956b18090e   1         6 minutes ago
```

smarterclayton:
Adding do not merge until we sort through API and naming semantics.

pkg/image/apis/image/types.go
```diff
@@ -308,6 +308,16 @@ type TagEventList struct {
 	Conditions []TagEventCondition
 }
 
+// NamedTagEventList relates a tag to its image history.
+type NamedTagEventList struct {
+	// Tag is the tag for which the history is recorded
+	Tag string
```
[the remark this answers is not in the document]
smarterclayton: To keep the struct exactly the same so we don't have drift.  The decision to use maps internally complicates this - this is the same struct as in v1 just in internal, but having the struct be the same keeps the validation logic close to external and makes it easier to reason about what is happening.

pkg/image/apis/image/types.go
```diff
@@ -408,6 +418,47 @@ type ImageStreamTagList struct {
...
+	Spec *TagReference
+	// status is the status tag details associated with this image stream tag, and it
+	// may be null if no push or import has been performed.
+	Status *NamedTagEventList
+	// image is the details of the most recent image stream status tag, and it may be
+	// null if import has not completed or an administrator has deleted the image
+	// object. To verify this is the most recent image, you must verify the generation
+	// of the most recent status.items entry matches the spec tag (if a spec tag is
+	// set).
+	Image *Image
```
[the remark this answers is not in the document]
smarterclayton: Why won't it work well with manifest lists?  Until we design support for that I would say "consistency with our existing APIs is first".

We have to maintain backwards compatibility with all the existing APIs anyway, so those APIs are probably going to look like the following (streams, tags, etc)

```
// This is the default image for the current platform architecture.  If there is no image for the current architecture, this is the first image in the manifest list.
Image *Image
// These are additional images available for other architectures.
AdditionalImages []Image
```

pkg/image/apiserver/registry/imagetag/rest.go
```diff
@@ -0,0 +1,427 @@
...
+			return nil, false, err
+		}
+		imageStream = &imageapi.ImageStream{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: namespace,
+				Name:      name,
+			},
+		}
+		rest.FillObjectMetaSystemFields(&imageStream.ObjectMeta)
+		create = true
```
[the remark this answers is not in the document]
smarterclayton: In Kube we're considering making all resources createable by update (since that's what PUT semantics offer).  However, this is to be consistent with ImageStreamTags and is trying to be the smallest delta from tags with the best possible API design for the future.

pkg/image/apis/image/types.go
```diff
@@ -408,6 +418,47 @@ type ImageStreamTagList struct {
...
+// A create operation will succeed if no spec tag has already been defined and the
+// spec field is set. Delete will remove both spec and status elements from the
+// image stream.
+type ImageTag struct {
+	metav1.TypeMeta
+	metav1.ObjectMeta
+
+	// spec is the spec tag associated with this image stream tag, and it may be null
+	// if only pushes have occurred to this image stream.
+	Spec *TagReference
```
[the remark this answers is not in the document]
smarterclayton: We have annotations on image stream tags so that the console can have metadata on them, and tooling can use them.  Our release tooling uses it.  So we can't change this anyway.

Difference between ObjectMeta.Name is that it is `foo:bar` and Spec.Name is `bar`.

pkg/image/apis/image/validation/validation.go
```diff
@@ -515,6 +516,76 @@ func ValidateImageStreamTagUpdateWithWhitelister(
...
+		result = append(result, ValidateImageStreamTagReference(whitelister, isRepositoryInsecure(newITag), *newITag.Spec, field.NewPath("spec"))...)
+	}
+
+	// ensure that only spec has changed
+	newISTCopy := *newITag
+	oldISTCopy := *oldITag
+	newISTCopy.Spec, oldISTCopy.Spec = nil, nil
+	newISTCopy.Image = oldISTCopy.Image
+	if !kapihelper.Semantic.Equalities.DeepEqual(&newISTCopy, &oldISTCopy) {
+		result = append(result, field.Invalid(field.NewPath("metadata"), "", "may not update fields other than spec"))
```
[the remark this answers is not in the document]
smarterclayton: yeah, will change.

pkg/image/apiserver/registry/imagetag/rest.go
```diff
@@ -0,0 +1,427 @@
...
+	itag, ok := obj.(*imageapi.ImageTag)
+	if !ok {
+		return nil, false, kapierrors.NewBadRequest(fmt.Sprintf("obj is not an ImageTag: %#v", obj))
+	}
+
+	// check for conflict
+	switch {
+	case len(itag.ResourceVersion) == 0:
+		// should disallow blind PUT, but this was previously supported
+		itag.ResourceVersion = imageStream.ResourceVersion
```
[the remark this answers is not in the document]
smarterclayton: Yeah, should have been a TODO.  Will go back and assess it again.

pkg/image/apis/image/types.go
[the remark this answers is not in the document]
dmage: Spec.Name can be computed from ObjectMeta.Name and therefore it is redundant. We don't have the image stream name there, why do we need to have the tag name?

ImageTag.ObjectMeta.Annotations -- I'd expect to see annotations of the tag there, not the image stream. The same about generation. Apparently that's how ImageStreamTags work today, is there any problems with the existing approach for annotations?

smarterclayton:
> The biggest question: why do we need imagestreamtag/imagetag virtual resources at all? We should remove either ImageStreams or Image(Stream)Tags, why do we need both?

We will likely never have sufficient reason to significantly change image streams (or image stream tags, really) - they allow atomic multi tag operations on a repository today, which we leverage in our release and CI testing.  The cost to change their structure aside, they are effectively mature, and our API policy really doesn't give us room to dramatically underpin them.  We also can't just remove ImageStreamTags - that would be too impactful to customers until some far future 5.0 release.

However ImageStreamTags definitely are both useful but suffer from some early mistakes (such as not including status).  This sets us up to improve the utility of single tag interactions (see the new CLI, which I think is a significant usability improvement) without a more complicated v2 version of ImageStreamTags (which would probably require we rev all v1 objects).  I tried to pick a name that was even simpler - if you care about tags, work with ImageTags.  If you care about streams, work with streams.  But we're so far into their lifecycle that replacement / significant refactor just won't benefit us or end users sufficiently.

pkg/image/apis/image/types.go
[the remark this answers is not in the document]
smarterclayton: Yeah, you couldn't view labels or annotations on the image streams from that resource, which was limiting for things like release controller and the command line tools.  Without the labels we can't select or group the streams, and while currently the virtual resources are providing a more ergonomic action on tags, they are still streams under the covers.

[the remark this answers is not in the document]
smarterclayton: I prefer to use the same struct in both places so that generic client code that deals with one or the other works the same.  But furthermore, you can't change the tag anyway and we default it on creation, so there should be no common interaction where a user *has* to set it.

pkg/image/apiserver/registry/imagetag/rest.go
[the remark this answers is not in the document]
smarterclayton: Ok, so this comment predates the point in time where we got comfortable with blind PUT.  Blind PUT allows you as a client to say "I don't care what the value is, the tag should be X" which is what we wanted for ImageStreamTags and I think we want to continue.  The tags command can simply send an ImageTag with spec set.

smarterclayton:
To clarify, my preferred goal with this resource is to improve ImageStreamTags and make it more useful for both display (see CLI output) and interaction, with room for future improvement.  It is not intended to endorse the more dramatic changes (removing IS or removing a v1 api), merely to better align the reality of user interaction and improve day to day for tools like release controller, oc tag, and UI ecosystem interactions on tags.

dmage:
@smarterclayton thanks for the explanation, looks good to me.

@adambkaplan @bparees ^^

smarterclayton:
If there are any review comments on the API, let me know.  I guess I'm waiting for Ben or Adam to have a chance to internalize the use case, at which point I think we can get the API in and add more e2e tests to this.

pkg/image/apiserver/registry/imagetag/rest.go
```diff
@@ -0,0 +1,427 @@
...
+	imageStreams, err := r.imageStreamRegistry.ListImageStreams(ctx, options)
+	if err != nil {
+		return nil, err
+	}
+
+	matcher := MatchImageTag(apihelpers.InternalListOptionsToSelectors(options))
+
+	list := &imageapi.ImageTagList{}
+	for _, currIS := range imageStreams.Items {
+		for currTag := range currIS.Status.Tags {
```
bparees: i guess i haven't internalized the use case....I thought one of the purposes of ImageTag was that it was going to return results for tags that only existed in imagestream.spec (ie had not been imported or pushed), as compared with ImageStreamTag which does not return those entries in list/404s on GET?

smarterclayton: yes, it will

bparees: this doesn't look like it will do that?

smarterclayton: Yikes, good catch.  I think we'll need to fetch images from the cache as well.

smarterclayton: Ok, I remember this context.  We didn't fetch images for imagestreamtags LIST because we didn't want to do an O(N) lookup of point images.  If even one person does `oc get imagestreamtags -A` we'd have to individually load each image.  We said at the time "if we need it we can add it" and I think that applies here, so I'm going to document for now that like image stream tags image tags don't retrieve image when LISTing and we can come back later and assess it.

bparees: hm.  I didn't follow what this has to do w/ fetching images, i just meant "if you don't find the tag in status, shouldn't you also iterate the tags in spec and return what you can from there if it exists"?  to avoid returning a 404 in the case of an imagestreamtag that has not yet successfully imported, or not yet been pushed(but for some reason the spec tag was created).

smarterclayton: I fixed the issue you reported (status and spec tags being reported).  That raised a second question - behavior of images (if the image does not exist).  In image stream tags the LIST imagestreamtags call does not retrieve images, and so I expanded this PR to specifically document that expectation in the code and the justification.

bparees: ah. i missed the push that fixed the status issue.

with respect to list behavior, i'm not seeing the documentation you say you added?  there's no godoc on the List function:
https://github.com/openshift/openshift-apiserver/pull/51/files#diff-6e2fba928b6d6601c9e8cd47e697892bR84

and nothing in the type godoc discusses list behavior or indicates that the image field is not populated for List operations:
https://github.com/openshift/openshift-apiserver/pull/51/files#diff-61b0127bad704c2776ee4eaec26e6de1R425

smarterclayton: Godoc is in the api PR (since I’ll just be revendoring it anyway)

smarterclayton: https://github.com/openshift/api/pull/523

bparees: got it, thanks

bparees:
@smarterclayton nothing further from me

smarterclayton:
Ok, do you want to review and approve api PR and then I’ll rebase here and start an e2e PR

pkg/image/apiserver/registry/imagetag/rest.go
```diff
@@ -0,0 +1,439 @@
...
+		imageStream.Spec.Tags[tag] = tagRef
+	} else {
+		// TODO: should error if it doesn't already exist?
+		delete(imageStream.Spec.Tags, tag)
+	}
+
+	// mutate the image stream
+	var newImageStream *imageapi.ImageStream
+	if create {
+		newImageStream, err = r.imageStreamRegistry.CreateImageStream(ctx, imageStream, &metav1.CreateOptions{})
```
deads2k: Direct registry calls bypass the safety of authorization, admission, and strategy prepare/normalize/validate.  We've eliminated direct registry calls from every other piece of the API surface.  Were I in @bparees and/or @dmage's shoes, I'm not sure I'd want to be the last man standing with this kind of pattern.  It's extremely hard to keep up to date and the risks to correctness, consistency, and safety are significant during every update.

smarterclayton: This is no different than stream tags, so while I can share your concerns in the abstract I don’t share them in the specific.

smarterclayton: Strategy is the same code, authorization is deliberately split (although in practice we allow), and admission is the big gap.  Closing admission at some point is going to be a good idea.

smarterclayton: Actually we call into all strategy behaviors by invoking the rest endpoint.  Where do you see differently (ie prepare and validation)?

smarterclayton:
e2e is https://github.com/openshift/origin/pull/24422 and must merge first @bparees

bparees:
@smarterclayton unhold when the origin change merges

smarterclayton:
Rebased, code is in origin for test, should pass.

smarterclayton:
Origin issues resolved, tagging

PR #53, opened 2019-12-03
Wire generic API server authentication and authorization flags

csrwng:
Exposes generic API server authentication and authorization flags to CLI so that they can be used by a hosted deployment of the openshift API server.

### DIFF
--- a/pkg/cmd/openshift-apiserver/cmd.go
+++ b/pkg/cmd/openshift-apiserver/cmd.go
@@ -9,11 +9,14 @@ import (
 	"path"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	genericapiserveroptions "k8s.io/apiserver/pkg/server/options"
 	"k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/klog"
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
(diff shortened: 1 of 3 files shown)


csrwng:
Fixes breakage (for hosted deployments) introduced by https://github.com/openshift/openshift-apiserver/pull/47

pkg/cmd/openshift-apiserver/cmd.go
```diff
@@ -37,6 +38,8 @@ var longDescription = templates.LongDesc(`
 
 func NewOpenShiftAPIServerCommand(name string, out, errout io.Writer, stopCh <-chan struct{}) *cobra.Command {
 	options := &OpenShiftAPIServer{Output: out}
+	authenticationOptions := genericapiserveroptions.NewDelegatingAuthenticationOptions()
```
sttts: shouldn't these be embedded into OpenShiftAPIServer ? That's usually the style we wire option structs.

sttts: And for that give `OpenShiftAPIServer` an AddFlags func which calls down into authn/z options.

pkg/cmd/openshift-apiserver/cmd.go
```diff
@@ -49,7 +52,7 @@ func NewOpenShiftAPIServerCommand(name string, out, errout io.Writer, stopCh <-c
 
 			serviceability.StartProfiler()
 
-			if err := options.RunAPIServer(stopCh); err != nil {
+			if err := options.RunAPIServer(authenticationOptions, authorizationOptions, stopCh); err != nil {
```
sttts: call `Validate` for auth options.

csrwng:
@sttts thx for the quick review. I've updated the PR

pkg/cmd/openshift-apiserver/cmd.go
```diff
@@ -30,13 +33,20 @@ import (
 type OpenShiftAPIServer struct {
 	ConfigFile string
 	Output     io.Writer
+
+	AuthenticationOptions *genericapiserveroptions.DelegatingAuthenticationOptions
```
sttts: drop Options postfix

csrwng: fixed

PR #61, opened 2019-12-13
reverts 1.17

deads2k:
it's messy and there are likely to be side-effects but it's a thing to try for impact before going further.


deads2k:
/hold

it can't merge like this

PR #65, opened 2019-12-19
Use IPv6 in resolver cache tests

adambkaplan:
Use "::1/128" instead of 127.0.0.1

### DIFF
--- a/pkg/image/apiserver/registryhostname/environmentresolvercache_test.go
+++ b/pkg/image/apiserver/registryhostname/environmentresolvercache_test.go
@@ -57,13 +57,13 @@ func TestServiceResolverCache(t *testing.T) {
 	c = fakeRetriever{
 		service: &corev1.Service{
 			Spec: corev1.ServiceSpec{
-				ClusterIP: "127.0.0.1",
+				ClusterIP: "::1",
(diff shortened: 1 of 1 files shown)


pkg/image/apiserver/registryhostname/environmentresolvercache_test.go
```diff
@@ -57,13 +57,13 @@ func TestServiceResolverCache(t *testing.T) {
 	c = fakeRetriever{
 		service: &corev1.Service{
 			Spec: corev1.ServiceSpec{
-				ClusterIP: "127.0.0.1",
+				ClusterIP: "::1/128",
```
dmage: @adambkaplan why do you need `/128` here? I wouldn't expect to see CIDR here.

adambkaplan: @dmage you're right, should just be `::1`. Make me wonder how this is passing (perhaps the /128 is ignored?)

adambkaplan: @dmage fixed - looks like the function under test just accepts the string as-is, doesn't do any verification

PR #69, opened 2020-01-27
Bug 1788700: include ISI error message changes when determining new generation need

gabemontero:
This change stems from a problem determination session that @dmage , @sdodson , and I had with a disconnected customer recently, and the difficulty in getting at the root cause of the customer's problem (which was a config error on their part).

The flow was:
- sample stream imports failed as expected going to registry.redhat.io
- customer changed the sample registry config to a mirrored repo, but the necessary config so that the image registry could communicate with that mirrored repo was not in place
- since the type/status/reason did not change with the new import error, the status condition was not bumped and when everyone looked at the imagestream status conditions, it looked like the sample registry config change did not take hold
- where if the message was considered, it would have bumped
- we had to run manual imports to learn of the config problem

This change fixes that (and lines up with the intent in the method comment above).

/assign @bparees 
(@adambkaplan is on PTO today)

### DIFF
--- a/pkg/image/apiserver/registry/imagestreamimport/rest.go
+++ b/pkg/image/apiserver/registry/imagestreamimport/rest.go
@@ -448,7 +448,7 @@ func checkImportFailure(status imageapi.ImageImportStatus, stream *imageapi.Imag
 		}
 	}
 
-	if !internalimageutil.HasTagCondition(stream, tag, condition) {
+	if !hasTagCondition(stream, tag, condition) {
 		setTagConditions(stream, tag, condition)
 		if tagRef, ok := stream.Spec.Tags[tag]; ok {
 			zero := int64(0)
@@ -457,6 +457,22 @@ func checkImportFailure(status imageapi.ImageImportStatus, stream *imageapi.Imag
 		}
 	}
 	return true
+}
+
+// hasTagCondition is an alternative to the public HasTagCondition in the internalimageutil package of this repo;
+// unlike that public method, it factors in changes to the message of the image stream import in case image stream
+// config changes (like which registry is used) still lead to an import InternalError. See
+// https://bugzilla.redhat.com/show_bug.cgi?id=1788700
+// So this returns true if the specified image stream tag has a condition with the same type, status, reason, and
+// message (but still does not check generation or date).
+func hasTagCondition(stream *imageapi.ImageStream, tag string, condition imageapi.TagEventCondition) bool {
+	for _, existing := range stream.Status.Tags[tag].Conditions {
+		if condition.Type == existing.Type && condition.Status == existing.Status && condition.Reason == existing.Reason &&
+			condition.Message == existing.Message {
+			return true
+		}
+	}
+	return false
 }
 
 // SetTagConditions applies the specified conditions to the status of the given tag.


pkg/image/apiserver/internalimageutil/helpers.go
```diff
@@ -471,7 +471,8 @@ func ResolveImageID(stream *imageapi.ImageStream, imageID string) (*imageapi.Tag
 // reason (does not check generation, date, or message).
 func HasTagCondition(stream *imageapi.ImageStream, tag string, condition imageapi.TagEventCondition) bool {
 	for _, existing := range stream.Status.Tags[tag].Conditions {
-		if condition.Type == existing.Type && condition.Status == existing.Status && condition.Reason == existing.Reason {
+		if condition.Type == existing.Type && condition.Status == existing.Status && condition.Reason == existing.Reason &&
+			condition.Message == existing.Message {
```
bparees: please update the godoc, since it currently explicitly says it does *not* check the message

dmage: As I can see, we have only one consumer, so I'd suggest to move this function to the consumer's package and make it private. It'll answer most of @bparees's questions:

2. if something relies on this function, we'll notice it,
3. that's exactly the reason why we need this change, we want see the up-to-date message for InternalErrors.

gabemontero: oops missed that in my scan re: godoc ... will address based on how the exchange below is reconciled.

there is only one call to this method in this repo and it is in https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/registry/imagestreamimport/rest.go#L451

I'm trying to figure out github's search repo's under github.com/openshift for confirmation outside this repo ... will report back when I have it working

Of if I can't I'll change the call in https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/registry/imagestreamimport/rest.go#L451 to a private version

re: you timestamp concern ... do you have a suggestion as to an alternative that mitigates that concern but addresses the debug pain we had @bparees ?

Based on how all the above goes, I'll rename the bugzilla or cancel it accordingly.

bparees: > re: you timestamp concern ... do you have a suggestion as to an alternative that mitigates that concern but addresses the debug pain we had @bparees ?

my alternative was to introduce a new function that is only used in places where we're confident we won't experience "flapping" due to changes in the message that were previously properly ignored.

but if indeed it's only used in that one spot there is not a concern that you'll be continuously updating something as a result of this change in behavior, then it's fine... but as @dmage suggested, making it private is a good idea if you're the only consumer.  (pretty much the same as introducing a new function, effectively).

gabemontero: I made https://github.com/openshift/openshift-apiserver/pull/69#discussion_r371329609 before seeing @dmage 's https://github.com/openshift/openshift-apiserver/pull/69#discussion_r371328366 :-)

gabemontero: ok I was treating the location/visibility of the method and potential flapping as separate concerns

if we agree we can combine them great 

i'l move on the private method path

gabemontero:
new private method introduced @dmage @bparees 

left existing method in place in case other repo's are using it (however remote that might be)

bugzilla is retitled

gabemontero:
bump @dmage for lgtm

gabemontero:
/hold

e2e-cmd may be failing in a consistent way here .... hard to compare with other PRs right now

may need to try this branch and e2e-cmd locally

gabemontero:
well the builds.sh test-cmd passes for me locally when running with these changes in my dev cluster 

I may need to push a debug commit to the PR to fully rule out this change.  

Then, based on the results, either address in the change or debug builds.sh

dmage:
from https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_openshift-apiserver/69/pull-ci-openshift-openshift-apiserver-master-e2e-cmd/192
```
FAILURE after 1.000s: test/cmd/builds.sh:41: executing 'oc new-build -D $'FROM centos:7' --to-docker' expecting success: the command returned the wrong error code
Standard output from the command:
warning: Cannot find git. Ensure that it is installed and in your path. Git is required to work with git repositories.
```

@gabemontero this doesn't seem to be related to this PR.

gabemontero:
yep that is what I've been referring to @dmage 

I would agree on it is not related, but I have not see this anywhere else.  And it has happened a few times in a row here.

If the current run fails, as I noted in https://github.com/openshift/openshift-apiserver/pull/69#issuecomment-579828760 I'll push a debug commit to confirm that it is not related, and then move on to debugging this as a test flake (where I'll probably have to add some cleanup to builds.sh)

gabemontero:
also @dmage the git thing is a red herring.

Here is why it actually fails:

```
 Standard error from the command:
    error: imagestreamtag.image.openshift.io "centos:7" already exists
[ERROR] hack/lib/cmd.sh:241: `return "${return_code}"` exited with status 1.
```

again, not sure why this change would cause it, but need to confirm one way or the other.

gabemontero:
anyway  that already exists results in

```
 Running test/cmd/builds.sh:41: executing 'oc new-build -D $'FROM centos:7' --to-docker' expecting success...
FAILURE after 1.000s: test/cmd/builds.sh:41: executing 'oc new-build -D $'FROM centos:7' --to-docker' expecting success: the command returned the wrong error code
```

gabemontero:
ok @dmage the e2e-cmd test is still running but I see from the project's test container:

```
started: (1/27/42) "[Suite:openshift/test-cmd][Serial][Disruptive] test-cmd: [Top Level] [Suite:openshift/test-cmd][Serial][Disruptive] test-cmd: test/cmd/builds.sh"
passed: (1m32s) 2020-01-29T17:54:01 "[Suite:openshift/test-cmd][Serial][Disruptive] test-cmd: [Top Level] [Suite:openshift/test-cmd][Serial][Disruptive] test-cmd: test/cmd/builds.sh"
```

so are flake notions have gained traction ... assuming that is right, I may still re-run it a few times to make sure and/or temporarily add the debug anyway to further make sure

gabemontero:
ok  this time builds.sh passed but images.sh failed with a `Get https://api.ci-op-kmqqmdr1-de4b5.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions?allowWatchBookmarks=true&fieldSelector=metadata.name%3Dversion&resourceVersion=26850&timeout=9m32s&timeoutSeconds=572&watch=true: dial tcp 54.148.114.170:6443: connect: connection refused`

perhaps related to the api.ci docker outage recently noted

PR #70, opened 2020-01-28
Bug 1795705: increase error logging around apiserver binary build flow

gabemontero:

/assign @bparees

### DIFF
--- a/pkg/build/apiserver/registry/buildconfiginstantiate/rest.go
+++ b/pkg/build/apiserver/registry/buildconfiginstantiate/rest.go
@@ -204,12 +204,13 @@ func (h *binaryInstantiateHandler) handle(r io.Reader) (runtime.Object, error) {
 					}
 				}
 			}
-			klog.V(2).Infof("failed to instantiate: %#v", request)
+			klog.V(2).Infof("failed to instantiate %#v with error %v", request, err)
 			return false, err
 		}
 		build = result
 		return true, nil
 	}); err != nil {
+		klog.Warningf("giving up trying to instantiate %#v due to: %v", request, err)
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


bparees:
let's add some logging to the switch statement that follows this:
https://github.com/openshift/openshift-apiserver/pull/70/files#diff-5ab90e4c1f4fada7caf2a5967999ff35L227

so that if we do fail to wait for the running build, we have a log output of what went wrong/why.

I *think* we already get this information back to the client pretty successfully, but it'd still be great to have the correlation if we need it.

bparees:
we should log here too so we know if we gave up on the instantiation attempt:
https://github.com/openshift/openshift-apiserver/pull/70/files#diff-5ab90e4c1f4fada7caf2a5967999ff35L213

gabemontero:
additional logs added @bparees

pkg/build/apiserver/registry/buildconfiginstantiate/rest.go
```diff
@@ -204,12 +204,13 @@ func (h *binaryInstantiateHandler) handle(r io.Reader) (runtime.Object, error) {
...
 				}
 			}
-			klog.V(2).Infof("failed to instantiate: %#v", request)
+			klog.V(2).Infof("failed to instantiate %#v with error %v", request, err)
 			return false, err
 		}
 		build = result
 		return true, nil
 	}); err != nil {
+		klog.Warningf("giving up trying to instantiate %v", request)
```
bparees:
```suggestion
		klog.Warningf("giving up trying to instantiate %#v due to: %v", request,err)
```

gabemontero: this update incorporated @bparees

PR #72, opened 2020-02-12
Bug 1777337: employ k8s label value validation when creating build pod build label…

gabemontero:
… value

Turns out we need label validation in addition to the existing prune label's length logic in the apiserver, like we did on OCM via https://github.com/openshift/openshift-controller-manager/pull/62

The original test scenario was from our CI cluster, which creates builds objs directly vs. instantiating them from the build config.

This change handles the instantiate from build config scneraio

@openshift/openshift-team-developer-experience

### DIFF
--- a/pkg/build/apiserver/buildgenerator/generator_test.go
+++ b/pkg/build/apiserver/buildgenerator/generator_test.go
@@ -914,10 +914,12 @@ func TestGenerateBuildFromConfig(t *testing.T) {
 	strategy := mockDockerStrategyForDockerImage(originalImage, &metav1.GetOptions{})
 	output := MockOutput()
 	resources := mockResources()
+	expectedLabel := "test-build-config-4"
 	bc := &buildv1.BuildConfig{
 		ObjectMeta: metav1.ObjectMeta{
-			UID:       "test-uid",
-			Name:      "test-build-config",
+			UID: "test-uid",
+			// Specify a name here that we cannot use as-is for a k8s ObjectMeta label
+			Name:      "test-build-config-4.3.0.ipv6-2019-11-27-0001-build",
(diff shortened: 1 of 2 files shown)


gabemontero:
non-build related failures in e2e-aws

/test e2e-aws

gabemontero:
bump @adambkaplan we have green tests PTAL

pkg/build/apiserver/buildgenerator/generator_test.go
```diff
@@ -916,8 +916,9 @@ func TestGenerateBuildFromConfig(t *testing.T) {
 	resources := mockResources()
 	bc := &buildv1.BuildConfig{
 		ObjectMeta: metav1.ObjectMeta{
-			UID:       "test-uid",
-			Name:      "test-build-config",
+			UID: "test-uid",
+			// Specify a name here that we cannot use as-is for a k8s ObjectMeta label
+			Name:      "test-build-config-4.3.0.ipv6-2019-11-27-0001-build",
```
dmage: is the expected label for this name `test-build-config-4`?

adambkaplan: I'm with @dmage - the test code should have an explicit expected value so it's clear we are testing that the generated labels are valid.

The alternative is to write a unit test just for `labelValue()`

gabemontero: which is what I did in the OCM version of this

I admittedly took a short cut here.

Will update in a sec.

gabemontero: updated @dmage @adambkaplan ptal thanks

gabemontero:
install readiness TO for e2e-aws-serial

PR #73, opened 2020-02-13
Bug 1801095: add all supported versions for pod mutators 

soltysh:
/assign @mfojtik 

Picks https://github.com/openshift/apiserver-library-go/pull/23


soltysh:
/hold
for more verification from @tnozicka

soltysh:
This is not needed here, we'll get it in the next bump. 
/close

PR #78, opened 2020-02-24
bug 1806557: stop creating SCC resources in openshift-apiserver

deads2k:
the kas-o placed these in the CVO manifests for management.


deads2k:
terraform
/retest

PR #83, opened 2020-03-13
Bug 1816534: DEVEXP-424: Leveraging node credentials during image stream import

ricardomaraschini:
With this PR we start to use node's pull secrets during image stream import process as described on the [enhancement](https://github.com/openshift/enhancements/pull/136) proposal. We prioritize namespace secrets over node pull secrets.

Note: Please review  https://github.com/openshift/cluster-openshift-apiserver-operator/pull/284 as well.


sttts:
nit: no dot at end of commit messages

pkg/image/apiserver/importer/importer_test.go
```diff
@@ -22,16 +22,18 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kapi "k8s.io/kubernetes/pkg/apis/core"
 
-	"github.com/openshift/library-go/pkg/image/registryclient"
+	imageref "github.com/openshift/library-go/pkg/image/reference"
 	imageapi "github.com/openshift/openshift-apiserver/pkg/image/apis/image"
 	dockerregistry "github.com/openshift/openshift-apiserver/pkg/image/apiserver/importer/dockerv1client"
 	"github.com/openshift/openshift-apiserver/pkg/image/apiserver/sysregistriesv2"
+
+	_ "github.com/openshift/openshift-apiserver/pkg/api/install"
```
sttts: why this line?

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -216,15 +214,14 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
-	// only load secrets if we need them
-	credentials := importer.NewLazyCredentialsForSecrets(func() ([]corev1.Secret, error) {
-		secrets, err := r.isV1Client.ImageStreams(namespace).Secrets(isi.Name, metav1.GetOptions{})
-		if err != nil {
-			return nil, err
-		}
-		return secrets.Items, nil
-	})
-	importCtx := registryclient.NewContext(r.transport, r.insecureTransport).WithCredentials(credentials)
+	secretsList, err := r.isV1Client.ImageStreams(namespace).Secrets(isi.Name, metav1.GetOptions{})
```
sttts: why is non-lazy loading ok when it wasn't before?

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -258,24 +257,6 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 		}
 	}
 
-	// if we encountered an error loading credentials and any images could not be retrieved with an access
```
sttts: this code was there to get proper error messages from the API. How is this done now?

pkg/image/apiserver/importer/importer_test.go
ricardomaraschini: We need to register some API types to run tests successfully, more specifically where we test serializing/deserializing:

```
converting (dockerpre012.DockerImage).Time.wall to (docker10.DockerImage).Time.wall: Cannot set dest. (Tried to deep copy something with unexported fields?)
```

That being said this import could be replaced by the following one(more restrict):

```
 "github.com/openshift/openshift-apiserver/pkg/user/apis/user/install"
```

The import on this commit was present in one of the deleted files so I opted to keep it as it was. If there is a better way or if this approach does not make sense at all please let me know.

sttts: `github.com/openshift/openshift-apiserver/pkg/user/apis/user/install` goes away soon with oauth-apiserver (will move to https://github.com/openshift/oauth-apiserver/tree/master/pkg/user/apis/user/install). This is not good. We cannot depend on user internal types, only `openshift/api/user`.

pkg/image/apiserver/registry/imagestreamimport/rest.go
ricardomaraschini: This is part of a behavioral change. Now if we can't load the secrets we return an API error to the user and don't create any image streams. With that out of the way(error retrieving secrets) future `StatusReasonUnauthorized` and `StatusReasonForbidden` would be correct so no need to overwrite them by `Unable to load secrets for this image` error.

Let me know what you reckon and I will adjust appropriately.

pkg/image/apiserver/registry/imagestreamimport/rest.go
ricardomaraschini: Please see https://github.com/openshift/openshift-apiserver/pull/83#discussion_r394857942

pkg/image/apiserver/importer/importer_test.go
ricardomaraschini: Sorry if I did not understand you correctly but I have refactored this to exclude the spurious import, PTAL. 

TL&DR:
- Imported `userapi github.com/openshift/api/user/v1` instead
- Running `userapi.Install(legacyscheme.Scheme)` in test's `init()`

pkg/image/apiserver/importer/importer_test.go
```diff
@@ -20,18 +20,25 @@ import (
...
-	"github.com/openshift/library-go/pkg/image/registryclient"
+	userapi "github.com/openshift/api/user/v1"
+	imageref "github.com/openshift/library-go/pkg/image/reference"
 	imageapi "github.com/openshift/openshift-apiserver/pkg/image/apis/image"
 	dockerregistry "github.com/openshift/openshift-apiserver/pkg/image/apiserver/importer/dockerv1client"
 	"github.com/openshift/openshift-apiserver/pkg/image/apiserver/sysregistriesv2"
 )
 
+func init() {
+	runtime.Must(userapi.Install(legacyscheme.Scheme))
```
sttts: why is the legacy scheme used? Installing anything into that is suspicious. Just create a new scheme object and install into it what you need.

pkg/image/apiserver/importer/staticcredentials.go
```diff
@@ -0,0 +1,81 @@
...
+
+	var cred auth.CredentialStore = registryclient.NoCredentials
+	if auths, found := keyring.Lookup(ref.String()); found {
+		cred = dockerregistry.NewStaticCredentialStore(&types.AuthConfig{
+			Username: auths[0].Username,
+			Password: auths[0].Password,
+		})
+	}
+
+	importCtx := registryclient.NewContext(
```
dmage: By recreating Context all the time, you are losing all caching features. Essentially you'll make a lot of ping requests to the /v2/ endpoint for every tag in ImageStreamImport.

pkg/image/apiserver/registry/imagestreamimport/rest.go
dmage: It was because of registries that doesn't need authentication:
```
$ curl -v 'https://registry.access.redhat.com/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/1.1 200 OK
```

But I'd say it's an exception
```
$ curl -v 'https://quay.io/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/2 401
< www-authenticate: Bearer realm="https://quay.io/v2/auth",service="quay.io"

$ curl -v 'https://registry-1.docker.io/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/1.1 401 Unauthorized
< Www-Authenticate: Bearer realm="https://auth.docker.io/token",service="registry.docker.io"

$ curl -v 'https://registry.redhat.io/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Bearer realm="https://registry.redhat.io/auth/realms/rhcc/protocol/redhat-docker-v2/auth",service="docker-registry"

$ curl -v 'https://gcr.io/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/2 401
< www-authenticate: Bearer realm="https://gcr.io/v2/token",service="gcr.io"

$ curl -v 'https://docker.bintray.io/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Bearer realm="https://docker.bintray.io/v2/token",service="docker.bintray.io",scope="repository:jfrog/reg2":pull,push

$ curl -v 'https://registry.svc.ci.openshift.org/v2/' 2>&1 | grep -i -e '^< HTTP' -e 'WWW-Authenticate'
< HTTP/1.1 401 Unauthorized
< Www-Authenticate: Bearer realm="https://registry.svc.ci.openshift.org/openshift/token"
```
and it makes our code much more complicated.

pkg/image/apiserver/importer/importer_test.go
ricardomaraschini: Functions[1] such as `schema1ToImage` or `schema2ToImage` on this package rely on the `legacyscheme`, not sure what the options were when they got developed. I don't mind at all fixing this to use a new scheme but it goes over the scope of this feature and I would address it as technical debt. If you are OK with leaving this as is for now I will create a new front to address this issue. You may even see TODO comments[2] talking about the `legacyscheme`. 

[1] https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/importer/image.go#L88
[2] https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/importer/dockerv1client/client.go#L34

sttts: ok, fine for me as a follow-up.

pkg/image/apiserver/importer/staticcredentials.go
ricardomaraschini: I can lay down a cache layer here and reuse them, what do you think? Ideally I would add this logic on `library-go` but unfortunately i can't as it includes k8s keyrings and I am not supposed to add these dependencies there.

ricardomaraschini: Could you PTAL ?

dmage: getting credentials using `ref` and then caching them by `regurl` definitely needs to be covered by tests. It can cause authentication problems.

pkg/image/apiserver/importer/importer.go
```diff
@@ -273,7 +273,7 @@ func (i *ImageStreamImporter) importFromRepository(ctx gocontext.Context, retrie
 
 	key := repositoryKey{url: *registryURL, name: repoName}
 	repo := &importRepository{
-		Ref:         ref,
+		Ref:         defaultRef,
```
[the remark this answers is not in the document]
ricardomaraschini: As far as I checked the only side effect is some error strings,  see [here](https://github.com/openshift/openshift-apiserver/pull/83/commits/40374369f09194564b6a801135bb199e2c5339c1#diff-8b6c07a981603ade97081700c13a9936R346) for an example. 

What changes is that:

1. If no registry is present, `docker.io` is used.
2. If no `Namespace` is set on the registry being imported and the repository is docker hub, `library` is used on the repository URL.
3. If no tag is set, `latest` is used.

Do you see any problem with this approach? Let me know if you have an use case where this can cause issues and I will make sure it works.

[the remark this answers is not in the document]
dmage: If it's for error strings, then it should contain the name as it was entered by the user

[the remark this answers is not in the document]
ricardomaraschini: Yes, you are right. I have reverted this change and now the error string is the same as it used to be. Please let me know if it is OK for you.

dmage:
Looks good, we'll need to create BZs to track follow-up tech debt addressing.

/lgtm

ricardomaraschini:
> Looks good, we'll need to create BZs to track follow-up tech debt addressing.
> 
> /lgtm

I have already opened a JIRA card for the tech debt, thanks for the heads-up.

pkg/image/apiserver/importer/staticcredentials.go
ricardomaraschini: /hold 

I am investigating what is the best way of achieving this.

ricardomaraschini: @dmage I have updated the indexing to use the repository name instead of the URL. Could you PTAL?

dmage:
/retitle Bug 1816534: DEVEXP-424: Leveraging node credentials during image stream import

dmage:
@mfojtik @sttts please take a look

PR #85, opened 2020-04-01
Use kubeconfig file provided in configuration when specified

irozzo-1A:
This PR is supposed to fixe the error I'm getting when running the openshift apiserver with an external `kubeConfig` file specified in the configuration [configuration](https://github.com/openshift/api/blob/master/config/v1/types.go#L268) . i.e.

```
F0401 17:48:59.903276       1 cmd.go:72] failed to get delegated authentication kubeconfig: failed to get delegated authentication kubeconfig: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
```

### DIFF
--- a/pkg/cmd/openshift-apiserver/cmd.go
+++ b/pkg/cmd/openshift-apiserver/cmd.go
@@ -123,6 +123,15 @@ func (o *OpenShiftAPIServer) RunAPIServer(stopCh <-chan struct{}) error {
 	configFileLocation := path.Dir(absoluteConfigFile)
 
 	config := obj.(*openshiftcontrolplanev1.OpenShiftAPIServerConfig)
+	// If authentication-kubeconfig or authorization-kubeconfig flag are not
+	// specified fallback to the kubeconfig file provided with the
+	// configuration.
+	if o.Authentication.RemoteKubeConfigFile == "" {
+		o.Authentication.RemoteKubeConfigFile = config.KubeClientConfig.KubeConfig
+	}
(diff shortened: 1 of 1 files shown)


irozzo-1A:
Hi @mfojtik, could you please check if this small fix makes sense to you? Thanks

stlaz:
Why don't you instead use the `--{authentication,authorization}-kubeconfig` flags?

irozzo-1A:
@stlaz I do actually, it's a minor stuff but I had the regression when migrating from `4.1` to `4.3` and I thought that it could be fixed without too much effort.

stlaz:
I can see that we previously were using this approach, but we use the in-cluster client config discovery, which couldn't be performed with your patch.

I think that if you want to prevent the in-cluster config search, you should specifically set the flags as you're doing today.

irozzo-1A:
> I can see that we previously were using this approach, but we use the in-cluster client config discovery now, which couldn't be performed with your patch.

IMHO my patch should have no impact for the in-cluster usage, because AFAIK in this case the [kubeConfig](https://github.com/openshift/api/blob/master/config/v1/types.go#L267) parameter should not be specified  (see https://github.com/openshift/kubernetes-apiserver/blob/master/pkg/server/options/authentication.go#L376). 

> I think that if you want to prevent the in-cluster config search, you should specifically set the flags as you're doing today.

Why specifying the `kubeConfig` in the configuration should not be enough to prevent in-cluster config search?

PR #91, opened 2020-04-14
Bug 1821771: new annotation needed for tightening cluster role bindings

vareti:
Adding a new annotation that can be used to reconcile RBAC resources created by openshift-apiserver. Used by [openshift-apiserver operator](https://github.com/openshift/cluster-openshift-apiserver-operator/pull/350) to remove unauthenticated users from list of subjects of some specific cluster role bindings

Needed for https://github.com/openshift/cluster-openshift-apiserver-operator/pull/350

### DIFF
--- a/pkg/bootstrappolicy/policy.go
+++ b/pkg/bootstrappolicy/policy.go
@@ -118,6 +118,9 @@ var (
 
 	// openShiftDescription is a common, optional annotation that stores the description for a resource.
 	openShiftDescription = "openshift.io/description"
+
+	// openshiftManagedRBACAnnotation is added to only RBAC resources created by openshift-apiserver.
+	openShiftManagedRBACAnnotation = "openshiftapiserver.rbac.authorization.openshift.io/managed"
 )
 
 func GetOpenshiftBootstrapClusterRoles() []rbacv1.ClusterRole {
@@ -856,6 +859,11 @@ func addDefaultMetadata(obj runtime.Object) {
 	for k, v := range bootstrappolicy.Annotation {
 		annotations[k] = v
 	}
+
+	// Setting the annotation to "true" would let github.com/openshift/cluster-openshift-apiserver-operator
+	// to manage this resource.
+	annotations[openShiftManagedRBACAnnotation] = "true"
+
 	metadata.SetAnnotations(annotations)
 }
 
--- a/test/testdata/bootstrappolicy/bootstrap_cluster_role_bindings.yaml
+++ b/test/testdata/bootstrappolicy/bootstrap_cluster_role_bindings.yaml
@@ -4,6 +4,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: system:masters
@@ -19,6 +20,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: system:node-admins
@@ -37,6 +39,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: cluster-admins
@@ -55,6 +58,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: cluster-readers
@@ -70,6 +74,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: basic-users
@@ -85,6 +90,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: self-access-reviewers
@@ -103,6 +109,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: self-provisioners
@@ -118,6 +125,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: system:oauth-token-deleters
@@ -136,6 +144,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: cluster-status-binding
@@ -151,6 +160,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: system:node-proxiers
@@ -166,6 +176,7 @@ items:
   kind: ClusterRoleBinding
   metadata:
     annotations:
+      openshiftapiserver.rbac.authorization.openshift.io/managed: "true"
       rbac.authorization.kubernetes.io/autoupdate: "true"
     creationTimestamp: null
     name: system:sdn-readers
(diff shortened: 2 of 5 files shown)


pkg/bootstrappolicy/policy.go
```diff
@@ -118,6 +118,10 @@ var (
 
 	// openShiftDescription is a common, optional annotation that stores the description for a resource.
 	openShiftDescription = "openshift.io/description"
+
+	// openShiftManagedRoleBindingAnnotation is a cluster role binding annotation used by
+	// openshift apiserver operator to manage it when set to true
+	openShiftManagedRoleBindingAnnotation = "rbac.authorization.openshift.io/managed"
```
sttts, line 124: is this a new one? If yes, we might want to make it clear to belong to openshift-apiserver? We also get oauth-apiserver and maybe we move RBAC rules eventually.

vareti: @sttts Yes, this is a new one. Does `openshiftapiserver.rbac.authorization.openshift.io/managed` looks clear?

vareti:
looks like a valid failure. looking into it.

vareti:
ci flakes

/retest

pkg/bootstrappolicy/policy.go
```diff
@@ -869,6 +873,11 @@ func addDefaultMetadata(obj runtime.Object) {
 	for k, v := range bootstrappolicy.Annotation {
 		annotations[k] = v
 	}
+
+	// Bug 1821771: We needed a way to tighten the cluster role
+	// binding when cluster is upgraded from 4.1
```
stlaz, line 878: Not a big fan of leaving BZ numbers through the code, the explanation at the annotation key definition should suffice

sttts:
> This annotation will be used by openshift apiserver operator when tightening cluster role bindings

Describe how it is used.

pkg/bootstrappolicy/policy.go
```diff
@@ -118,6 +118,9 @@ var (
 
 	// openShiftDescription is a common, optional annotation that stores the description for a resource.
 	openShiftDescription = "openshift.io/description"
+
+	// annotation used to indicate if reconciliation is needed on a RBAC resource
```
sttts, line 122: what does this mean? Who reconciles what?

vareti: updated the comment. PTAL.

pkg/bootstrappolicy/policy.go
```diff
@@ -118,6 +118,11 @@ var (
 
 	// openShiftDescription is a common, optional annotation that stores the description for a resource.
 	openShiftDescription = "openshift.io/description"
+
+	// openshiftManagedRBACAnnotation is added to only RBAC resources created by openshift-apiserver
```
marun, line 122: Grammar nit for code comments - if longer than a phrase, format as sentences (leading word capitalized, period, etc).

```
// This is a phrase

// This is a sentence. That way there is clear separation between it and the next sentence. 
```

I think it preferable to avoid documenting the specifics here of how the operator will respond to the annotation. Instead, point to the operator repo if more detail is required. That way, if/when the operator behavior needs to change, a future maintainer will not be at risk of not updating the comment in the operand.

pkg/bootstrappolicy/policy.go
```diff
@@ -856,6 +861,12 @@ func addDefaultMetadata(obj runtime.Object) {
 	for k, v := range bootstrappolicy.Annotation {
 		annotations[k] = v
 	}
+
+	// setting the annotation to true would let openshift-apiserver
```
marun, line 865: nit: multi-line comment - format as a sentence.

pkg/bootstrappolicy/policy.go
vareti: That sounds correct and gives us less maintainability.  I updated the comment now.

pkg/bootstrappolicy/policy.go
vareti: Done

vareti:
@marun Thanks for the review. I addressed your comments. Could you please take a look again.

marun:
@vareti The annotation addition proposed here would appear to only apply to ClusterRoleBindings created on bootstrap. In the case of an existing cluster, how is the annotation going to be set? And if it won't be set, doesn't that imply that the annotation should not be set by the operand, and that the absence of the annotation is an indication that unauthorized subject removal should be performed?  e.g. `rbac.authorization.openshift.io/unmanaged=true` would be set by a user to indicate that they didn't want to have the unauthorized subject to be removed from the ClusterRolebinding?

vareti:
> In the case of an existing cluster, how is the annotation going to be set?

If the annotation was missing, it will be added on restart. But if the annotation is modified, no changes will be made and retains the modified value upon openshift-apiserver restart.So, In case of an upgrade from 4.1 or user deleting the annotation, it would be added back whenever openshift-apiserver is restarted.

If a user want to indicate that the unauthorized subject should not be removed, only way it could be done is by setting `openshiftapiserver.rbac.authorization.openshift.io/managed=false`. This behavior is not guaranteed if the annotation was deleted.

The idea is to have separate annotation that can be used with RBAC resources created by openshift-apiserver. This is similar to kube-apiserver having a separate annotation for the RBAC resources that it creates.

We can achieve the same intent using the behavior that you suggested also. But following upstream behavior would be less confusing I guess?

pkg/bootstrappolicy/policy.go
```diff
@@ -856,6 +859,11 @@ func addDefaultMetadata(obj runtime.Object) {
 	for k, v := range bootstrappolicy.Annotation {
 		annotations[k] = v
 	}
+
+	// Setting the annotation to "true" would let github.com/openshift/cluster-openshift-apiserver-operator
+	// to manage this resource.
+	annotations[openShiftManagedRBACAnnotation] = "true"
```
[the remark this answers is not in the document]
vareti: I could not find an easy way to add the annotation just to cluster role bindings without duplicating code. Thought that there is no harm in adding annotation using this method.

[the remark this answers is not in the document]
marun: Not sure what code duplication you're talking about. You have the metadata. Only apply the annotation when kind+api group indicate a ClusterRoleBinding.

sttts:
> If a user want to indicate that the unauthorized subject should not be removed, only way it could be done is by setting openshiftapiserver.rbac.authorization.openshift.io/managed=false. This behavior is not guaranteed if the annotation was deleted.

We briefly discussed this in Slack I think. My question was: is this a new feature that we allow customers to opt-out of reconciliation? We should be super careful to get this right. What does upstream do if an object exists, but is modified? What did we do in openshift-apiserver?

pkg/bootstrappolicy/policy.go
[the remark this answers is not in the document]
vareti: Yes. That's a good point. Let me do that change.

vareti:
> is this a new feature that we allow customers to opt-out of reconciliation? 

Yes. We are giving the ability to opt-out of reconciliation. @deads2k suggested to add this ability previously [here](https://github.com/openshift/cluster-openshift-apiserver-operator/pull/350#discussion_r408167585). This follows the upstream behavior of [auto-reconciliaton](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation). 

When I deleted openshift RBAC annotation and restart openshift-apiserver, I saw that the annotation was added back. I did not test deleting upstream RBAC annotation. I will test this and confirm soon.

sttts:
> Yes. We are giving the ability to opt-out of reconciliation. @deads2k suggested to add this ability previously [here](https://github.com/openshift/cluster-openshift-apiserver-operator/pull/350#discussion_r408167585). This follows the upstream behavior of [auto-reconciliaton](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation).
> 
> When I deleted openshift RBAC annotation and restart openshift-apiserver, I saw that the annotation was added back. I did not test deleting upstream RBAC annotation. I will test this and confirm soon.

Shouldn't we more tightly follow what upstream does in https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation? I.e. even use their label.

vareti:
@sttts +1 for reusing the same label again. I tried to follow whatever David commented. But using upstream label is better if I think about it. That will align us more tightly with upstream. If it is the case, this PR is not needed anymore.

sttts:
> @sttts +1 for reusing the same label again

Better call it `rbac.authorization.openshift.io/autoupdate` to not confuse the two controllers. But same semantics if possible.

vareti:
> Better call it rbac.authorization.openshift.io/autoupdate to not confuse the two controllers

@sttts if we are having a separate annotation, do you think it should indicate the component? Like you said in one of your previous [comment](https://github.com/openshift/openshift-apiserver/pull/91#discussion_r408933761)

sttts:
> @sttts if we are having a separate annotation, do you think it should indicate the component? Like you said in one of your previous [comment](https://github.com/openshift/openshift-apiserver/pull/91#discussion_r408933761)

If we use the same annotation, won't kube-apiserver wipe that resource?

vareti:
> If we use the same annotation, won't kube-apiserver wipe that resource?

Today, the same upstream annotation is already added to bootstrap policies created by the openshift-apiserver.

```
oc get clusterrolebinding system:openshift:discovery -o jsonpath="{.metadata.annotations}"
map[rbac.authorization.kubernetes.io/autoupdate:true]
```

With this, my understanding is that, at start-up openshift-apiserver RBAC resources are reconciled the same way kube-apiserver reconciles its resources. So, having this annotation would not wipe the resource, I guess?

And we can have the controller in the openshift-apiserver operator to remove unauthenticated subjects from specific cluster role bindings.

pkg/bootstrappolicy/policy.go
```diff
@@ -118,6 +118,9 @@ var (
 
 	// openShiftDescription is a common, optional annotation that stores the description for a resource.
 	openShiftDescription = "openshift.io/description"
+
+	// openshiftManagedRBACAnnotation is added to only RBAC resources created by openshift-apiserver.
```
sttts, line 122:
```
// openshiftManagedRBACAnnotation is added to *all* RBAC resources created by openshift-apiserver.
// This way, openshift-apiserver-operator will know which RBAC resources are owned by openshift-apiserver
// and which are not, e.g. in order to update them on upgrades.
```

pkg/bootstrappolicy/policy.go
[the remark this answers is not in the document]
sttts: This then contradicts the description that all RBAC resources are annotated. Be more specific above.

sttts:
/close

We switched to a doc approach as changing this automatically might break existing clusters.

PR #102, opened 2020-04-30
Bug 1731222: unwarranted SARs on foreground GC of objects with owner refs

gabemontero:
See https://coreos.slack.com/archives/CB48XQ4KZ/p1588008887095400 for history

/assign @deads2k 

@deads2k - I could not find any official guideline on how to (a presumably temporary until a rebase pulls it in) copy of k8s in lieu of vendoring; please advise if code changes / commit wording etc. need modification

/assign @dmage 

@dmage if possible minimally pls make sure this change catches your instance of the problem

@openshift/openshift-team-developer-experience @bparees  FYI

/hold

since the upstream PR https://github.com/kubernetes/kubernetes/pull/90534  has not merged yet ... however, if during review we think it is just a matter of when I'll unhold

### DIFF
--- a/pkg/build/apiserver/admission/strategyrestrictions/admission.go
+++ b/pkg/build/apiserver/admission/strategyrestrictions/admission.go
@@ -17,7 +17,6 @@ import (
 	authorizationclient "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/client-go/rest"
 	kapihelper "k8s.io/kubernetes/pkg/apis/core/helper"
-	rbacregistry "k8s.io/kubernetes/pkg/registry/rbac"
 
 	"github.com/openshift/api/build"
 	buildclient "github.com/openshift/client-go/build/clientset/versioned"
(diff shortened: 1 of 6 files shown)


gabemontero:
note for reviewers - I've confirmed this changes is not needed prior to 4.5, at least for templates 

I confirmed all the way back to 4.2

gabemontero:
> Nit on the commit & PR title:
> 
>     1. Reference the bug in the title line of the git commit
> 
>     2. Leave a shorter title that makes sense to a customer
> 
> 
> You should continue to reference the upstream PR in the details of the commit.

both PR title and commit updated @adambkaplan

gabemontero:
all green tests again @deads2k @dmage @adambkaplan

deads2k:
/hold

move the package to where you use it please

gabemontero:
per slack conversation with @deads2k we put versions of the new mutating GC check in both packages

deads2k:
/lgtm
/hold cancel

this duplication makes sense because these are logically distinct resources that are unlikely to live together long term.  The upstream reference and the reflective test ensure they will stay in sync.  This also eliminates a k/k dep.  Good PR.

PR #104, opened 2020-05-06
Bug 1832762: updates Dockerfile.rhel to golang 1.13 and ocp:4.4

p0lyn0mial:


images/Dockerfile.rhel
```diff
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
```
tnozicka, line 1: you need to follow up with updating the CI config

p0lyn0mial: done - https://github.com/openshift/release/pull/8905

PR #108, opened 2020-05-13
Bug 1833558: create cluster roles for modiyfing scc-s

soltysh:
/assign @stlaz


soltysh:
Replaced with https://github.com/openshift/cluster-kube-apiserver-operator/pull/856

PR #110, opened 2020-05-21
Bug 1774616: Remove insecureSkipTLSVerifyBackend bypass for BuildLogOptions

coreydaley:
Remove the bypass for InsecureSkipTLSVerifyBackend when comparing
BuildLogOptions and PodLogOptions
Depends on https://github.com/openshift/api/pull/652


coreydaley:
/hold
Waiting on https://github.com/openshift/api/pull/652

pkg/build/apis/build/log_options_test.go
```diff
@@ -14,10 +14,6 @@ func TestLogOptionsDrift(t *testing.T) {
 	for i := 0; i < popts.NumField(); i++ {
 		// Verify name
 		name := popts.Field(i).Name
-		// TODO: Add this option to deployment config log options
```
bparees, line 17: i hope this comment was wrong and should have said buildconfig log options.

coreydaley: Looking at the rest of the file it is comparing BuildLogOptions and PodLogOptions, so I would hope so ;)

bparees:
@coreydaley is there no work that needs to be done to actually wire the build log request logic up to use the new field?

coreydaley:
> @coreydaley is there no work that needs to be done to actually wire the build log request logic up to use the new field?

Quite possibly, working on the rest of this pr now.

coreydaley:
@bparees ptal

bparees:
so the field is not used?  oc logs has a --insecure field, i'd expect that when that field is set and you're doing an "oc logs build/foo" invocation, the insecure field gets propagated into this new field?

bparees:
/hold

```
func BuildToPodLogOptions(opts *buildapi.BuildLogOptions) *kapi.PodLogOptions {
```

needs to be updated to map this field, i believe.

coreydaley:
@bparees the field has been updated

bparees:
as @adambkaplan noted on https://github.com/openshift/oc/pull/433, the vendor changes should be in a bump commit.

coreydaley:
@bparees fixed commit message

PR #111, opened 2020-06-02
IR-21: Leveraging image config for "insecure" registries during ImageStreamImport

ricardomaraschini:
By leveraging `images.config.openshift.io/cluster` "insecure" registry configuration we now can flag a registry as insecure for all image stream imports. Without this patch users had to set the registry as "insecure" on the ImageStreamImport request, regardless of what was defined on `images.config.openshift.io/cluster`.

I have chosen to include here some other housekeeping changes:

1. Standardized the receiver name for "ImageStreamImporter" struct (linter warning)
2. Migrating from `golang.org/x/net/context` to `context`
3. Removed unnecessary function parameters (using struct "properties" instead)
4. Vendoring [sysregistriesv2](https://github.com/containers/image/blob/master/pkg/sysregistriesv2/system_registries_v2.go) so we can remove a own copy of this package we had laying on our repository

All four changes above were constricted into their own commits (plus commits where we vendor stuff) in an attempt to make the review easier.

Tests for this feature were implemented through https://github.com/openshift/origin/pull/25058


pkg/image/apiserver/sysregistriesv2/system_registries_v2.go
```diff
@@ -116,6 +116,39 @@ func (config *V2RegistriesConf) Nonempty() bool {
 		len(config.UnqualifiedSearchRegistries) != 0)
 }
 
+// Insecure returns if access to registry at location is flagged as insecure, i.e. invalid
```
dmage, line 119: This implementation doesn't match the registries.conf documentation.

Please keep it in sync with https://github.com/containers/image/blob/master/pkg/sysregistriesv2/system_registries_v2.go

You may want to import this package directly, AFAIK they fixed their "types" package and it shouldn't have 100500 dependencies anymore.

ricardomaraschini: I did not know this package was copied from somewhere else. I have updated this PR to vendor the package and remove the copy&pasta we had.

pkg/image/apiserver/importer/importer.go
```diff
@@ -81,38 +89,68 @@ func NewImageStreamImporter(retriever RepositoryRetriever, regConf *sysregistrie
 		limiter:   limiter,
 		regConf:   regConf,
 
-		digestToRepositoryCache: make(map[gocontext.Context]map[manifestKey]*imageapi.Image),
+		digestToRepositoryCache: make(map[context.Context]map[manifestKey]*imageapi.Image),
```
sttts, line 92: that's interesting type. Which context ends up here? The one of the request?

pkg/image/apiserver/importer/importer_test.go
```diff
@@ -11,7 +12,7 @@ import (
 	"testing"
 	"time"
 
-	"golang.org/x/net/context"
+	"github.com/containers/image/pkg/sysregistriesv2"
```
sttts, line 16: nit: remove empty line

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -223,7 +237,7 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 		r.transport, r.insecureTransport, secretsList.Items,
 	)
 	imports := r.importFn(importCtx, v2regConf)
-	if err := imports.Import(ctx.(gocontext.Context), isi, stream); err != nil {
+	if err := imports.Import(ctx.(context.Context), isi, stream); err != nil {
```
sttts, line 240: why this case?

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -246,7 +260,7 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 			r.transport, r.insecureTransport, nil,
 		)
 		imports := r.importFn(importCtx, v2regConf)
-		if err := imports.Import(ctx.(gocontext.Context), isi, stream); err != nil {
+		if err := imports.Import(ctx.(context.Context), isi, stream); err != nil {
```
sttts, line 263: why this cast?

pkg/image/apiserver/registry/imagestreamimport/rest.go
ricardomaraschini: Nice catch, removed it.

pkg/image/apiserver/registry/imagestreamimport/rest.go
ricardomaraschini: Nice catch, removed it.

pkg/image/apiserver/importer/importer_test.go
ricardomaraschini: Removed.

pkg/image/apiserver/importer/importer.go
ricardomaraschini: I had the same impression when I came across this construction for the first time, seems rather interesting. I did some digging and I discovered that this is entangled with the client request's context. The next obvious question I had was: "what?!". I could not find where this was freed so I thought this would grow indefinitely, a real memory hoarder eating everything on its way. Well, this is not the case because the object holding it is created in a per request basis, so the next obvious question was "why?", if it is based on a per request basis why to index per context? My guess is that this object was planned to handle multiple requests at the same time but never got finished.

@dmage Am I missing something here? I think we could have a task to check(fix?) this as tech-debt, what do you think?

sttts: while we touch this, can you add a comment? This has big code smell.

dmage: I created a tech-debt card: https://issues.redhat.com/browse/IR-116

dmage:
@sttts can you take a look again?

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -205,9 +208,20 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 		klog.Warningf("failed to load ImageContentSourcePolicy resources, mirrored images will not be found: %v", err)
 	}
 
-	v2regConf := &sysregistriesv2.V2RegistriesConf{}
-	err = registries.EditRegistriesConfig(v2regConf, nil, nil, icspRules)
+	imageConfig, err := r.imageCfgV1Client.Images().Get(
+		context.TODO(), "cluster", metav1.GetOptions{},
```
sttts, line 212: why not use ctx?

ricardomaraschini: fixed.

ricardomaraschini:
@sttts @dmage Tests for this were passing OK but now we need a rebase (most likely due to some other vendor update merged recently). I am planning to move on with the rebase as soon as I get an approval that the changes here are correct, this way I hope to avoid a possible second rebase later down the line.

pkg/image/apiserver/importer/importer.go
```diff
@@ -87,32 +95,32 @@ func NewImageStreamImporter(retriever RepositoryRetriever, regConf *sysregistrie
 }
 
 // Import tries to complete the provided isi object with images loaded from remote registries.
-func (i *ImageStreamImporter) Import(ctx gocontext.Context, isi *imageapi.ImageStreamImport, stream *imageapi.ImageStream) error {
+func (iptr *ImageStreamImporter) Import(ctx gocontext.Context, isi *imageapi.ImageStreamImport, stream *imageapi.ImageStream) error {
```
sttts: nit: bad abbrevation starting with i, but the i does not match the first letter of the type. `imp` would be better IMO.

pkg/image/apiserver/importer/importer.go
```diff
@@ -929,3 +972,7 @@ func setImageImportStatus(images *imageapi.ImageStreamImport, i int, tag string,
 func invalidStatus(position string, errs ...*field.Error) metav1.Status {
 	return kapierrors.NewInvalid(schema.GroupKind{Group: "", Kind: ""}, position, errs).ErrStatus
 }
+
+func forbiddenStatus(err error) metav1.Status {
+	return kapierrors.NewForbidden(schema.GroupResource{Group: "", Resource: ""}, "", err).ErrStatus
```
sttts, line 977: nit: what's the value of this func?

pkg/image/apiserver/importer/importer.go
ricardomaraschini: fixed.

pkg/image/apiserver/importer/importer.go
ricardomaraschini: Individually, maybe none. It is there because I have seen a very similar one called `invalidStatus()` a few lines above.

dmage:
/override ci/prow/e2e-cmd
this test is broken for now

dmage:
@sttts can you override e2e-cmd? @soltysh is working on it, but it's broken for now.

petr-muller:
/hold

If the test is broken then there's no point in retesting. Either button merge it or /override after a failure is reported and then unhold (but Tide may decide to re-run the tests before merge so button merge is probably better).

sttts:
ci/prow/e2e-cmd

PR #112, opened 2020-06-02
Bug 1841035: Add RBAC rules for volume snapshots

jsafrane:
* Allow regular users to read VolumeSnapshotClasses (similar to StorageClass)
* Allow regular users to read/write VolumeSnapshots in their namespaces (similar to PVCs)
* Allow storage-admin to read/write VolumeSnapshotClasses and VolumeSnapshotContents (~ PV) and read VolumeSnapshots (~PVCs)

### DIFF
--- a/pkg/bootstrappolicy/policy.go
+++ b/pkg/bootstrappolicy/policy.go
@@ -60,6 +60,7 @@ var (
 	schedulingGroup            = "scheduling.k8s.io"
 	kAuthzGroup                = kauthorizationapi.GroupName
 	kAuthnGroup                = kauthenticationapi.GroupName
+	snapshotGroup              = "snapshot.storage.k8s.io"
 
 	deployGroup         = oapps.GroupName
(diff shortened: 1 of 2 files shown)


deads2k:
> * Allow regular users to read VolumeSnapshotClasses (similar to StorageClass)
> * Allow regular users to read/write VolumeSnapshots in their namespaces (similar to PVCs)
> * Allow storage-admin to read/write VolumeSnapshotClasses and VolumeSnapshotContents (~ PV) and read VolumeSnapshots (~PVCs)

These appear to be non-standard kube resources, probably added via CRD.  The proper way to  add a new resource is to add an aggregated clusterrole in the operator where you're adding the CRD itself: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

/hold

deads2k:
came up in arch call, I think it was clear

/close

PR #113, opened 2020-06-03
Bug 1841035: Modify basic-user and storage-admin to be aggregated

jsafrane:
We want to be able to add them permissions to read/write snapshots. Since volume snapshots are based on CRDs, those rules will be added in csi-snapshot-controller-operator. See https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/39

### DIFF
--- a/pkg/bootstrappolicy/policy.go
+++ b/pkg/bootstrappolicy/policy.go
@@ -290,8 +290,23 @@ func GetOpenshiftBootstrapClusterRoles() []rbacv1.ClusterRole {
 			},
 		},
 		{
+			// Aggregated: cluster-csi-snapshot-controller-operator needs to grant storage-admin permissions to manage
+			// VolumeSnapshotContents (=~PV) and VolumeSnapshotClasses (=~StorageClass).
 			ObjectMeta: metav1.ObjectMeta{
 				Name: StorageAdminRoleName,
+			},
+			AggregationRule: &rbacv1.AggregationRule{
+				ClusterRoleSelectors: []metav1.LabelSelector{
+					{
+						MatchLabels: map[string]string{"storage.openshift.io/aggregate-to-storage-admin": "true"},
+					},
+				},
+			},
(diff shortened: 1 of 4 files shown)


jsafrane:
@deads2k, together with https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/39 I get the roles / bindings I need.

But I had to make `basic-user` and `storage-admin` aggregated. Will upgrade from non-aggregated cluster work?

pkg/bootstrappolicy/policy.go
```diff
@@ -293,6 +293,19 @@ func GetOpenshiftBootstrapClusterRoles() []rbacv1.ClusterRole {
 			ObjectMeta: metav1.ObjectMeta{
 				Name: StorageAdminRoleName,
 			},
+			AggregationRule: &rbacv1.AggregationRule{
+				ClusterRoleSelectors: []metav1.LabelSelector{
+					{
+						MatchLabels: map[string]string{"rbac.authorization.k8s.io/aggregate-to-storage-admin": "true"},
```
deads2k, line 299: `storage.openshift.io/aggregate-to-storage-admin`

pkg/bootstrappolicy/policy.go
```diff
@@ -429,6 +442,19 @@ func GetOpenshiftBootstrapClusterRoles() []rbacv1.ClusterRole {
 					openShiftDescription: "A user that can get basic information about projects.",
 				},
 			},
+			AggregationRule: &rbacv1.AggregationRule{
+				ClusterRoleSelectors: []metav1.LabelSelector{
+					{
+						MatchLabels: map[string]string{"rbac.authorization.k8s.io/aggregate-to-basic-user": "true"},
```
deads2k, line 448: `authorization.openshift.io/aggregate-to-basic-user`

Also, are you planning to add something here?  This is a very broadly granted role.

PR #114, opened 2020-06-03
Bug 1843496: project: improve error message on cluster-scoped template objects

sttts:
Before:

> Internal error occurred: the server could not find the requested resource

After:

> Internal error occurred: user.openshift.io/v1, Kind=Group is not namespace-scoped


mfojtik:
/lgtm

(feel free to retag after resolving the conflict)

PR #115, opened 2020-06-04
Bug 1751258: Using manifest digest on manifest list import

ricardomaraschini:
When importing a manifest list only one manifest is imported therefore
we need to use this one manifest digest instead of the digest for the
whole manifest list when checking for integrity.


dmage:
@ricardomaraschini as it's a bug fix, it deserves a test

ricardomaraschini:
> @ricardomaraschini as it's a bug fix, it deserves a test

Well, I implemented an unit test to validate a ManifestList import. PTAL.

dmage:
@adambkaplan can you take a look?

PR #116, opened 2020-06-05
Bug 1844479: Add custom conversion for binary builds 

adambkaplan:
Adding a manual conversion func to create BinaryBuildRequestOptions from
url query parameters. This is an attempt to address some of the issues in #95 that prevent the 1.19 rebase.

Before I go down the rabbit hole of reflection APIs, I want to make sure that this is generally the right thing to do for this situation. The test case converts url query parameters to a `BinaryBuildRequestOptions`, which exists to ensure `oc start-build` works properly.

### DIFF
--- a/pkg/api/serialization_test.go
+++ b/pkg/api/serialization_test.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"fmt"
 	"math/rand"
+	"net/url"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/google/gofuzz"
+	fuzz "github.com/google/gofuzz"
 
 	"k8s.io/apimachinery/pkg/api/apitesting"
 	"k8s.io/apimachinery/pkg/api/apitesting/fuzzer"
@@ -22,7 +24,7 @@ import (
 	kapitesting "k8s.io/kubernetes/pkg/api/testing"
 	"k8s.io/kubernetes/pkg/apis/apps"
 	kapi "k8s.io/kubernetes/pkg/apis/core"
-	"k8s.io/kubernetes/pkg/apis/core/v1"
+	corev1 "k8s.io/kubernetes/pkg/apis/core/v1"
 	"k8s.io/kubernetes/pkg/apis/core/validation"
 	extensionsv1beta1 "k8s.io/kubernetes/pkg/apis/extensions/v1beta1"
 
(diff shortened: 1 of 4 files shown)


pkg/build/apis/build/v1/conversion.go
```diff
@@ -150,6 +152,24 @@ func Convert_build_BuildStrategy_To_v1_BuildStrategy(in *newer.BuildStrategy, ou
 	return nil
 }
 
+func Convert_url_Values_To_v1_BinaryBuildRequestOptions(in *url.Values, out *v1.BinaryBuildRequestOptions, s conversion.Scope) error {
+	if in == nil || out == nil {
+		return nil
+	}
+	out.AsFile = in.Get("asFile")
+	out.Commit = in.Get("revision.commit")
```
adambkaplan, line 160: This is where I intend to use reflection to do the inverse of `queryparams.Convert` [1]

[1] https://github.com/kubernetes/apimachinery/blob/master/pkg/conversion/queryparams/convert.go#L135

sttts: You know your target type. Why reflection?

adambkaplan:
/retest

Failed to acquire lease, and various flakes.

adambkaplan:
/retest

Unrelated upgrade flakes

adambkaplan:
/hold

Per @deads2k we need to ensure the conversions are working via a fuzz test (at least proof of a failure if we have a bad fuzz function). I tried to do this in #118, but to no avail.

adambkaplan:
@sttts perhaps you can help?

adambkaplan:
@sttts updated with the inverse conversion func.

#118 has the conversion fuzz test. You can see in the failure that the conversion function is invoked. `ObjectMeta` is not converted, but per the `queryparams` converter [1] that is not possible?

[1] https://github.com/kubernetes/apimachinery/blob/master/pkg/conversion/queryparams/convert.go#L132-L135

adambkaplan:
/retest

#118 has been updated with the fuzz test in what should be a passing/green state. Commits have been squashed for clarity.

deads2k:
Can you add the unit test from 118?  It should be passing now.  You can reset object meta to do your diff.

adambkaplan:
/hold cancel

@deads2k @sttts fuzzing unit test added (with only one iteration instead of 1000).

adambkaplan:
/retest

Unrelated flake: `[sig-scheduling] SchedulerPredicates [Serial] validates resource limits of pods that are allowed to run [Conformance] [Suite:openshift/conformance/serial/minimal] [Suite:k8s] `

pkg/api/serialization_test.go
```diff
@@ -22,7 +24,7 @@ import (
 	kapitesting "k8s.io/kubernetes/pkg/api/testing"
 	"k8s.io/kubernetes/pkg/apis/apps"
 	kapi "k8s.io/kubernetes/pkg/apis/core"
-	"k8s.io/kubernetes/pkg/apis/core/v1"
+	v1 "k8s.io/kubernetes/pkg/apis/core/v1"
```
deads2k, line 27: nit, we import this as corev1

deads2k:
test looks good

/approve

@adambkaplan I think someone on your team should own lgtm so they can see how this gets wired.

adambkaplan:
/assign @otaviof 

The conversion from query params to API objects is part of how binary builds get wired through to build pods today.

adambkaplan:
@otaviof merged your suggestion & squashed. PTAL

PR #121, opened 2020-06-26
IR-99: Adding metric for registry v1 protocol imports

ricardomaraschini:
This commit adds a new metric: a counter of attempts to import images from registries using protocol v1. This protocol version has been deprecated in the past but before we remove its support from apiserver we need to be sure nobody is using it.

A PrometheusRule will be created to alert on top of this counter, as we don't have any Prometheus rule on this repository it will be created on Image Registry operator project (we already have other rules related to Image Registries in there).

### DIFF
--- a/pkg/image/apiserver/importer/importer.go
+++ b/pkg/image/apiserver/importer/importer.go
@@ -794,6 +794,7 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
 	}
 
 	// load digests
+	imported := false
 	for i := range repository.Digests {
 		importDigest := &repository.Digests[i]
 		if importDigest.Err != nil || importDigest.Image != nil {
@@ -811,6 +812,7 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
 			importDigest.Err = err
 			continue
 		}
+		imported = true
 	}
 
 	for i := range repository.Tags {
@@ -830,6 +832,11 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
 			importTag.Err = err
 			continue
 		}
+		imported = true
+	}
+
+	if imported {
+		v1ImageImportsCounter.WithLabelValues(repository.Ref.String()).Inc()
 	}
 }
 
--- a/pkg/image/apiserver/importer/metrics.go
+++ b/pkg/image/apiserver/importer/metrics.go
@@ -1,0 +1,21 @@
+package importer
+
+import (
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var (
+	v1ImageImportsCounter = k8smetrics.NewCounterVec(
+		&k8smetrics.CounterOpts{
+			Subsystem: "apiserver",
+			Name:      "v1_image_imports_total",
+			Help:      "Counter of images imported using v1 protocol",
+		},
+		[]string{"repository"},
+	)
+)
+
+func init() {
+	legacyregistry.MustRegister(v1ImageImportsCounter)
+}


pkg/image/apiserver/importer/importer.go
```diff
@@ -671,6 +671,8 @@ func (isi *ImageStreamImporter) importRepositoryFromDocker(ctx gocontext.Context
 		if err != nil {
 			klog.V(5).Infof("unable to access repository %#v: %#v", repository, err)
 			if strings.HasSuffix(err.Error(), "does not support v2 API") {
+				v1ImageImportsCounter.Inc()
+				klog.Warningf("attempting to import image %q using v1 registry, this is going to be deprecated.", repository.Ref.String())
```
dmage, line 675:
```
apiVersion: image.openshift.io/v1
kind: ImageStream
metadata:
  name: badis
spec:
  dockerImageRepository: google.com/foo/bar
```

This imagestream will increment the counter, but google.com is not a registry and it doesn't support v1 API.

ricardomaraschini: Fixed. PTAL.

pkg/image/apiserver/importer/importer.go
```diff
@@ -830,6 +832,12 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
 			importTag.Err = err
 			continue
 		}
+		imported = true
+	}
+
+	if imported {
+		v1ImageImportsCounter.Inc()
+		klog.Warningf("attempting to import image %q using v1 registry, this is going to be deprecated.", repository.Ref.String())
```
dmage, line 840: This feature is deprecated and might be removed in a future release

ricardomaraschini: done.

pkg/image/apiserver/importer/importer.go
```diff
@@ -830,6 +832,12 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
 			importTag.Err = err
 			continue
 		}
+		imported = true
+	}
+
+	if imported {
+		v1ImageImportsCounter.Inc()
```
deads2k, line 839: will we want to label which repository it came from in some way?  I would expect the cardinality on that to be reasonable.

pkg/image/apiserver/importer/importer.go
```diff
@@ -830,6 +832,12 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
...
+		klog.Warningf("imported image %q using v1 registry, this feature is deprecated and might be removed in a future release", repository.Ref.String())
```
deads2k, line 840: this is likely to be too noisy in the logs.

deads2k:
/approve 
/hold

I'd like you to get a +1 from @openshift/openshift-team-monitoring and they can remove the hold.

Also, see the question about adding a label for repositories

pkg/image/apiserver/importer/importer.go
dmage: We want to give customers some clues about what uses the deprecated API. If we can label the metric with the repository name, we can remove this message.

pkg/image/apiserver/importer/metrics.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var (
+	v1ImageImportsCounter = k8smetrics.NewCounter(&k8smetrics.CounterOpts{
+		Subsystem: "apiserver",
+		Name:      "v1_image_imports",
```
simonpasquier, line 11: this should be `v1_image_imports_total` to follow [good practices](https://prometheus.io/docs/practices/naming/#metric-names).

pkg/image/apiserver/importer/importer.go
simonpasquier: How many repository values would you expect in practice?
One problem with having a `repository` label is that you might not know the label values in advance which means the metric will only be visible after it's been incremented once (see Prometheus doc about [missing metrics](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics)). Not knowing much about the project, I can't say if it's an issue here.

ricardomaraschini: We don't expect this to have many labels (in fact we expect it to have a total of none). This metric will be used solely for alerting, it is OK if it is not present.

pkg/image/apiserver/importer/metrics.go
ricardomaraschini: Fixed, PTAL.

pkg/image/apiserver/importer/importer.go
ricardomaraschini: Let me add some extra info to this: imports using registry protocol v1 are rare (or that is what we at least believe in), i.e. cardinality here is unknown but expected to be low. I have sent over a commit that adds the repository as label as proposed by @deads2k . PTAL.

To see the big picture: the alert on top of this metric is designed [here](https://github.com/openshift/cluster-image-registry-operator/pull/573/files#diff-993500c6c19512d419d76efd8a98015bR20-R37), I had to do some effort to take the initial `Inc()` into account, inspired by github.com/prometheus/prometheus/issues/1673

pkg/image/apiserver/importer/importer.go
ricardomaraschini: As a bonus for adding the repository as a label on the metric we can now get rid of this log line. Just did it.

pkg/image/apiserver/importer/importer.go
simonpasquier: Overall I'm concerned that the alert isn't really actionable by a cluster admin and triggers alert fatigue. At the very least I would expect the `severity` label to be `info` rather than `warning`. Also looking at the definition of the alert, it could be that the alert starts firing and then resolves itself because no import using protocol v1 happened in the last 30 minutes.

cc @brancz @lilic  for additional inputs.

ricardomaraschini: We are trying to deprecate the support for registry v1 protocol but for that to happen we need to first guarantee that no one is leveraging it, hence the alert. It is designed to solve itself after a while as you mentioned, the idea is to have the information available for us and for the client so we can let v1 go away peacefully. 

cc @dmage

dmage: I don't think it should be `info`, it's an actionable item. We don't expect anybody to use it, and if somebody somehow still use it, they should stop doing that. The feature is going to be removed in 4.7.

lilic: Agreed with Simon here, we had this problem before with another deprecated API alert before done by the workloads team, when both users and our OSD SREs were confused by it and it was too noisy, I would advise against this alert. 

I understand the easy use of alerts as notifications to our users, but this is exactly the symptom vs cause based alerts. https://prometheus.io/docs/practices/alerting/ 

Can this be mentioned in the release docs, that you are deprecating registry protocol v1? You yourself say its not that many people who use this in the first place?

dmage: @lilic if it's too noisy, the customer has a huge problem: they rely on something that will be gone.

It'll be mentioned in the release docs, but how customers will know that they actually use this feature?

dmage: Anyway, this PR is about adding the metrics, please discuss alert at https://github.com/openshift/cluster-image-registry-operator/pull/573

lilic: Sorry about that, only seen it after this, answered there :)

pkg/image/apiserver/importer/importer.go
```diff
@@ -830,6 +832,11 @@ func importRepositoryFromDockerV1(ctx gocontext.Context, repository *importRepos
 			importTag.Err = err
 			continue
 		}
+		imported = true
+	}
+
+	if imported {
+		v1ImageImportsCounter.WithLabelValues(repository.Ref.String()).Inc()
```
lilic, line 839: As Simon mentioned before in the PR review, how many possible values of repository could there be? Is this a bound number or a random generated string? 

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

See explanation on this here https://prometheus.io/docs/practices/naming/#labels

dmage: @lilic as @deads2k said, the cardinality on that is expected to be reasonable. We don't expect this feature to be used and therefore any labels to be created.

dmage:
@lilic do you want something to be changed or can we merge it?

dmage:
@lilic @simonpasquier can we merge it?

PR #127, opened 2020-08-03
Bug 1861696: Using DockerClientDefaults on block/insecure check

ricardomaraschini:
When user does not provide a registry docker.io is used by default. By
calling DockerClientDefaults() we assure that we always have it properly
set.


dmage:
@ricardomaraschini can you add a test for this?

ricardomaraschini:
> @ricardomaraschini can you add a test for this?

I can add one in `origin`

PR #143, opened 2020-09-30
Bug 1862532: Fix label value validation for BuildConfigs

coreydaley-redhat:
Label check should allow underscores, hyphens, and decimals in the name
    instead of restricting to DNS RFC 1123 compatible values


adambkaplan:
/retitle Bug 1862532: Fix label value validation for BuildConfigs

adambkaplan:
/hold

I think we need https://github.com/openshift/openshift-controller-manager/pull/137 to land first

adambkaplan (approved):
/lgtm

PR #154, opened 2020-11-03
Remove the code serving the oauth and user APIs

stlaz:
The code moved to the oauth-apiserver. Uservalidation used here is vendored from there for now.

/cc @sttts @p0lyn0mial


p0lyn0mial:
could we hold this PR until we change the operators so that `oas-o` won't claim these groups?

stlaz:
/hold
sure, I think we have to

stlaz:
/hold cancel
/retest
the OAS operator should no longer claim these APIs as of https://github.com/openshift/cluster-openshift-apiserver-operator/pull/417

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/openshift/build-machinery-go v0.0.0-20200917070002-f171684f77ab
 	github.com/openshift/client-go v0.0.0-20201123133249-10558821f936
 	github.com/openshift/library-go v0.0.0-20201102091359-c4fa0f5b3a08
+	github.com/openshift/oauth-apiserver v0.0.0-20201022180844-09d7ea02b5ae
```
sttts, line 32: surprsing dependency

stlaz: It turns out that OAS reused some uservalidation that moved along with the API to the other server. Maybe apiserver-library-go would be the right place to share this code?

https://github.com/openshift/openshift-apiserver/pull/154/commits/09845a13b74c6a4d6c60d656bed2870f8038b781#diff-d5e126898d69f39400e851ed5f5effbb63db8391c3492d8014f06d3eaba73585R13

sttts: How much is "some" ?

stlaz: It's all the additions in the https://github.com/openshift/openshift-apiserver/pull/154/commits/09845a13b74c6a4d6c60d656bed2870f8038b781 commit.

Namely:
- requestlimit admission (appears to be limiting the number of allowed projects?)
- the legacy openshift authorization API + its conversion and conversion testing
- serialization tests

sttts: Where is requestlimit admission used in oauth-apiserver?

> the legacy openshift authorization API + its conversion and conversion testing

used where?

sttts: Or in other words, which pieces could we change / split / duplicate to cut the dependency?

stlaz: > the legacy openshift authorization API + its conversion and conversion testing
    - `validation.ValidateUserName`, `validation.ValidateGroupName` in conversions to be able to convert the kind of the subjects, and it's used in validation to validate the same
> Where is requestlimit admission used in oauth-apiserver?
It's not, the username validation from the oauth-apiserver is used in requestlimit audition to determine whether the user requesting projects is a system user or openshift user

> Or in other words, which pieces could we change / split / duplicate to cut the dependency?
we could factor out username and groupname validation

sttts: In seriallization_test.go we can certainly get rid of ValidateUserName/ValidateGroupName completely by just using valid names (e.g. `validusername%d`) directly.

pkg/api/serialization_test.go
```diff
@@ -424,18 +423,6 @@ func originFuzzer(t *testing.T, seed int64) *fuzz.Fuzzer {
 			j.Spec.Template.Spec.InitContainers = nil
 			j.Status.Template.Spec.InitContainers = nil
 		},
-		func(j *oauthapi.OAuthAuthorizeToken, c fuzz.Continue) {
```
sttts, line 427: do we have the serialization_test in oauth-apiserver now?

PR #155, opened 2020-11-03
pin to github.com/openshift/kubernetes-client-go

p0lyn0mial:


p0lyn0mial:
the change was introduced in https://github.com/openshift/library-go/pull/926

PR #158, opened 2020-11-09
Use imageapi with custom v1.SecretList type

alvaroaleman:
Requires a rebase after https://github.com/openshift/client-go/pull/168 merges
/hold


vendor/github.com/openshift/api/image/v1/generated.pb.go
```diff
@@ -1070,6 +1098,7 @@ func init() {
 	proto.RegisterType((*NamedTagEventList)(nil), "github.com.openshift.api.image.v1.NamedTagEventList")
 	proto.RegisterType((*RepositoryImportSpec)(nil), "github.com.openshift.api.image.v1.RepositoryImportSpec")
 	proto.RegisterType((*RepositoryImportStatus)(nil), "github.com.openshift.api.image.v1.RepositoryImportStatus")
+	proto.RegisterType((*SecretList)(nil), "github.com.openshift.api.image.v1.SecretList")
```
sttts, line 1101: this makes me nervous. @smarterclayton will this string show up in the payload somehow such that an old client wouldn't know what to do with a `github.com.openshift.api.image.v1.SecretList`?

sttts: From the Decode function I think we are safe:
```
// Decode attempts to convert the provided data into a protobuf message, extract the stored schema kind, apply the provided default
// gvk, and then load that data into an object matching the desired schema kind or the provided into. If into is *runtime.Unknown,
// the raw data will be extracted and no decoding will be performed. If into is not registered with the typer, then the object will
// be straight decoded using normal protobuf unmarshalling (the MarshalTo interface). If into is provided and the original data is
// not fully qualified with kind/version/group, the type of the into will be used to alter the returned gvk. On success or most
// errors, the method will return the calculated schema kind.
```

sttts: In other words, only the group version kind matters to select the target object to unmarshal into. And as the proto messages for both core secret list and image secret list are the same, the decoding will work.

sttts:
On the serializer side, with https://github.com/openshift/openshift-apiserver/pull/158#discussion_r527800822 for proto, I think we are safe with the wrapper type.

@deads2k which concerns to you have?

PR #162, opened 2020-12-07
Bug 1905195: Detecting broken connections to the Kube API takes up to 15 minutes

p0lyn0mial:

ref: https://github.com/openshift/library-go/pull/945


rhpmali:
As per pre-merge verification. The bug is pre-merge verified in https://bugzilla.redhat.com/show_bug.cgi?id=1905195#c3 , adding QE lgtm:
/lgtm

crawford:
Cherry-pick-approved removed. Tests are too flaky and we already have too many fixes to be verified this cycle.

soltysh:
e2e-cmd is known to be flakey and it's currently awaiting openshift/origin#25831

mfojtik:
/override ci/prow/e2e-cmd

This is known broken and the fix just merged.

PR #163, opened 2020-12-08
Bug 1894918: Panic output due to timeouts in openshift-apiserver

p0lyn0mial:

PR #164, opened 2020-12-09
Bug 1905991: Detecting broken connections to the Kube API takes up to 15 minutes

p0lyn0mial:

ref: https://github.com/openshift/library-go/pull/967

Note:
Only the last 2 commits are relevant. Previous were added and reviewed https://github.com/openshift/openshift-apiserver/pull/163


knobunc:
[As patch-manager] Pulling the cherry-pick-approval since the PR didn't make the merge window... sorry. Please resolve the CI problems so it can make next week's window.

PR #165, opened 2020-12-10
Bug 1867380: When using webhooks in OCP 4.5 fails to rollout latest deploymentconfig

p0lyn0mial:
`DeploymentConfig` kind is registered under `v1` in the scheme. Registration happens here https://github.com/openshift/openshift-apiserver/blob/e75d326783d96501e0904dfd787eca4fc08b900a/pkg/apps/apis/apps/v1/register.go#L15

When calling a webhook we need to specify the version explicitly otherwise the kind won't be found in the registry and the requests will fail with `no kind "DeploymentConfig" is registered for version "apps.openshift.io/" in scheme`

It seems this was the only place it was broken. Feel free to double-check that.


sttts:
Great find!

sttts:
/lgtm
/approve

cc @adambkaplan

PR #167, opened 2020-12-14
Bug 1896977: Enhance API host name validation

candita:
Change route validation to apply the stricter validation rules from IsFullyQualifiedDomainName on host name when OpenShiftAllowNonDNSCompliantHostAnnotation is false or missing: host name must have at least two labels, with each label less than 64 chars, no disallowed characters, total length less than 254 chars, and trailing dots are allowed. If OpenShiftAllowNonDNSCompliantHostAnnotation is true then use the former lenient validation of just IsDNS1123Subdomain: total length less than 254, and no disallowed characters.

Add new annotation: OpenShiftAllowNonDNSCompliantHostAnnotation = "openshift.io/allow-non-dns-compliant-host" to allow configuations that suppress the stricter validation.  This may validate routes that cannot be admitted (same behavior as before).

Add/update unit tests.


candita:
Companion PR in openshift/router: https://github.com/openshift/router/pull/238

pkg/route/apis/route/validation/validation.go
```diff
@@ -27,10 +28,17 @@ func ValidateRoute(route *routeapi.Route) field.ErrorList {
...
-		if len(kvalidation.IsDNS1123Subdomain(route.Spec.Host)) != 0 {
-			result = append(result, field.Invalid(specPath.Child("host"), route.Spec.Host, "host must conform to DNS 952 subdomain conventions"))
+		if lenient, ok := route.Annotations[oapi.OpenShiftAllowNonDNSCompliantHostAnnotation]; ok && lenient == "true" {
+			if len(kvalidation.IsDNS1123Subdomain(route.Spec.Host)) != 0 {
+				result = append(result, field.Invalid(specPath.Child("host"), route.Spec.Host, "host must conform to DNS subdomain conventions"))
+			}
+		} else {
+			errs := kvalidation.IsFullyQualifiedDomainName(specPath.Child("host"), route.Spec.Host)
+			if len(errs) != 0 {
+				result = append(result, field.Invalid(specPath.Child("host"), route.Spec.Host, fmt.Sprintf("host must conform to DNS 1123 subdomain conventions: %v", errs)))
```
wking, line 40: I don't understand why the `else` branch's `IsFullyQualifiedDomainName` error gets associated with `DNS 1123 subdomain conventions` in the error message, while the previous `if` branch's `IsDNS1123Subdomain` error gets the generic `DNS subdomain conventions` conventions.  Wouldn't `IsDNS1123Subdomain` be the one looking for DNS 1123 subdomain violations?

pkg/route/apis/route/validation/validation_test.go
```diff
@@ -65,13 +66,75 @@ func TestValidateRoute(t *testing.T) {
...
 				},
+				Spec: routeapi.RouteSpec{
+					Host: "host.com",
+					To:   createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 1,
+		},
+		{
+			name: "Non-dns compliant host with override annotation",
```
wking, line 77: nit: `dns` -> `DNS` here and in other human-oriented strings.

pkg/route/apis/route/validation/validation_test.go
```diff
@@ -65,13 +66,75 @@ func TestValidateRoute(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "name",
 				},
+				Spec: routeapi.RouteSpec{
+					Host: "host.com",
+					To:   createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 1,
```
wking, line 74: probably out of scope for this PR, but I prefer tests for error generation to place some constraints on the error strings.  For an example, see [here][1], or, without `stretchr/testify`, [here][2].

[1]: https://github.com/openshift/installer/blob/ec982ecf5bd847add0f3af82b04f32251972701c/pkg/types/aws/validation/machinepool_test.go#L74-L78
[2]: https://github.com/openshift/cluster-version-operator/blob/d3e90d913bf2ebdcdccd2132f7d56fa29c2aaf3b/pkg/cvo/cvo_test.go#L3445-L3453

pkg/route/apis/route/validation/validation_test.go
```diff
@@ -65,13 +66,75 @@ func TestValidateRoute(t *testing.T) {
...
+				},
+				Spec: routeapi.RouteSpec{
+					Host: "host",
+					To:   createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 1,
+		},
+		{
+			name: "Generated host name too long",
```
wking, line 125: nit: `Spec.Host` is explicitly set here, so "Generated host name too long" -> "Specified host name too long"?  `Spec.Host` could have been set by default-host-generating logic, or could have been set by whoever created the Route in the first place, and this test-case doesn't need to distinguish between those two cases.

pkg/route/apis/route/validation/validation_test.go
wking: Somewhat in this space, because it's a bit wiggly without actual error string comparison: do we want explicit checks for "really long subdomain, but overall hostname is short enough" showing we error on that (or not, with the override annotation)?

pkg/route/apis/route/validation/validation.go
candita: `IsFullyQualifiedDomain` calls `IsDNS1123Subdomain`, `IsDNS1123Label`, and performs several other checks, making it truy conform to RFC1123.

pkg/route/apis/route/validation/validation_test.go
candita: We have a check for that in L132: https://github.com/openshift/openshift-apiserver/pull/167/files/2c5d8f04d03e980a5bb25b096e7d513ac7784381#diff-e571a492e8cca2251f9208e5274c4c9a4e67368113b1b6f131dd5b685c5eaa56R132

pkg/route/apis/route/validation/validation.go
```diff
@@ -27,10 +28,17 @@ func ValidateRoute(route *routeapi.Route) field.ErrorList {
 
 	specPath := field.NewPath("spec")
 
-	//host is not required but if it is set ensure it meets DNS requirements
+	// host is not required but if it is set ensure it meets certain DNS requirements
 	if len(route.Spec.Host) > 0 {
-		if len(kvalidation.IsDNS1123Subdomain(route.Spec.Host)) != 0 {
-			result = append(result, field.Invalid(specPath.Child("host"), route.Spec.Host, "host must conform to DNS 952 subdomain conventions"))
+		if lenient, ok := route.Annotations[oapi.OpenShiftAllowNonDNSCompliantHostAnnotation]; ok && lenient == "true" {
```
deads2k, line 33: you cannot tighten this for updates that are not changing the route.spec.host.  That means you'll need to find a way to split up the update validation.

deads2k:
as long as we don't break route REST updates and the only impact is to 
1. routes that don't work today
2. routes that use custom routers today that may accept names our router rejects and the NE team has good reason to believe this is a very narrow use-case

I think this reasonable.

pkg/route/apis/route/validation/validation.go
candita: Thanks @deads2k , that sounds reasonable.  I have made a change and updated some unit tests. PTAL when you get a chance.

candita:
https://bugzilla.redhat.com/show_bug.cgi?id=1895107 describes the issue with the failing e2e-cmd test.

candita:
Error on e2e-cmd may be due to https://bugzilla.redhat.com/show_bug.cgi?id=1916880

candita:
/test e2e-cmd

Try one last time before reporting issue.  Miciah pointed out other occurences of same issue in https://search.ci.openshift.org/?search=ClusterOperators+did+not+settle%3A+timed+out+waiting+for+the+condition&maxAge=48h&context=1&type=bug%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job

candita:
I was advised to ignore the failure in e2e-cmd, as it is not a Required test.

pkg/route/apiserver/registry/route/strategy.go
```diff
@@ -134,6 +134,7 @@ func (s routeStrategy) allocateHost(ctx context.Context, route *routeapi.Route)
 func (s routeStrategy) Validate(ctx context.Context, obj runtime.Object) field.ErrorList {
 	route := obj.(*routeapi.Route)
 	errs := s.allocateHost(ctx, route)
+	errs = append(errs, validation.ValidateHost(route)...)
 	errs = append(errs, validation.ValidateRoute(route)...)
```
Miciah, line 138: Is there a reason to have the `Validate` method in `strategy.go` call the `ValidateHost` function instead of having the `ValidateRoute` function call `ValidateHost`?  This change makes `Validate` and `ValidateUpdate` less consistent, and it also goes against the pattern whereby an API's `strategy.go` has `ValidateFoo` and `ValidateFooUpdate` methods that call the `ValidateFoo` and `ValidateFooUpdate` functions, respectively, that are defined in `validation.go` to do all the actual validation.  

(Granted, it's a little weird that the route's `Validate` method calls `allocateHost` to set a default `spec.host` instead of defining the defaulting in `pkg/route/apis/route/v1/defaults.go`, but that is different from validation.  It's also a little weird that `ValidateHostUpdate` calls `validateHostUpdate`, but `validateHostUpdate` is doing a permissions check, not exactly validation.)

candita: Yes, there is a reason.  @deads2k  asked me to split out the host validation so we only check hostname on an update if it was changed - see (https://github.com/openshift/openshift-apiserver/pull/167#discussion_r545338195)

Miciah: `ValidateRouteUpdate` has this check:  https://github.com/openshift/openshift-apiserver/blob/132f77c43a683412bb1e77e15385811e2f22c934/pkg/route/apis/route/validation/validation.go#L101-L103

Ah, but I see now what you mean:  `ValidateRouteUpdate` calls `ValidateRoute`.  This is a bit weird and confusing.  What do you think about renaming `ValidateRoute` to `ValidateRouteCreate` and adding a `validateRoute` method with the common logic?  So `strategy.go` would look like this:  
```go
func validateHost(route *routeapi.Route) field.ErrorList {
	// Same logic that ValidateHost has in this PR (but now the function can be private).
}

func validateRoute(route *routeapi.Route) field.ErrorList {
	// Common validation logic that both ValidateRouteCreate and ValidateRouteUpdate need.
}
func ValidateRouteCreate(route *routeapi.Route) field.ErrorList {
	// Call validateRoute and validateHost and return the results.
}

func ValidateRouteUpdate(route *routeapi.Route, older *routeapi.Route) field.ErrorList {
	// Same logic that it has in this PR.
}
```

Miciah:
/lgtm
The cleanups in https://github.com/openshift/openshift-apiserver/pull/167#discussion_r562928379 and https://github.com/openshift/openshift-apiserver/pull/167#discussion_r562958174 can be considered for a follow-up.

pkg/route/apiserver/registry/route/strategy.go
candita: Will be addressed in a follow-up.

pkg/route/apis/route/validation/validation.go
```diff
@@ -88,9 +82,25 @@ func ValidateRoute(route *routeapi.Route) field.ErrorList {
 	return result
 }
 
+// ValidateHost checks that a route's host name meets DNS requirements.
+func ValidateHost(route *routeapi.Route) field.ErrorList {
+	result := field.ErrorList{}
+	if len(route.Spec.Host) < 1 {
```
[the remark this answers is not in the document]
candita: Will be addressed in a follow-up.

pkg/route/apis/route/validation/validation.go
```diff
@@ -27,13 +28,6 @@ func ValidateRoute(route *routeapi.Route) field.ErrorList {
 
 	specPath := field.NewPath("spec")
 
-	//host is not required but if it is set ensure it meets DNS requirements
-	if len(route.Spec.Host) > 0 {
```
deads2k, line 31: This looks wrong.  A method called `ValidateRoute` should validate the route.

pkg/route/apis/route/validation/validation.go
```diff
@@ -88,9 +82,25 @@ func ValidateRoute(route *routeapi.Route) field.ErrorList {
...
+	return result
+}
+
 func ValidateRouteUpdate(route *routeapi.Route, older *routeapi.Route) field.ErrorList {
 	allErrs := validation.ValidateObjectMetaUpdate(&route.ObjectMeta, &older.ObjectMeta, field.NewPath("metadata"))
 	allErrs = append(allErrs, validation.ValidateImmutableField(route.Spec.WildcardPolicy, older.Spec.WildcardPolicy, field.NewPath("spec", "wildcardPolicy"))...)
+	if route.Spec.Host != older.Spec.Host {
+		allErrs = append(allErrs, ValidateHost(route)...)
+	}
 	allErrs = append(allErrs, ValidateRoute(route)...)
```
deads2k, line 104: This looks improper based on our previous conversation.  The rules for updating are different than the rules for creating.  This is a long-standing pattern in kube where we have three distinct methods with different rules

1. validateObj - this validates objects for create semantics
2. validateObjUpdate - this validates *differently* than validateObj because create calls can be tighter than update calls.  Also, this should ignore status.
3. validateObjectStatusUpdate - this validates differently than validateObjUpdate because status updates should never fail on valid specs.

Your validateRoute function should fully validate a route for the purposes of create.  Because the update semantics don't match the create semantics, you cannot call one from the other.  Making your validateRoute function not fully validate a route isn't a good idea.

pkg/route/apis/route/validation/validation.go
candita: My first attempt to split out the host validation didn't fit the model, so I will fix.

pkg/route/apis/route/validation/validation.go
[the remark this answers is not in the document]
candita: Done

pkg/route/apis/route/validation/validation.go
candita: I have addressed this, and Miciah's similar comment https://github.com/openshift/openshift-apiserver/pull/167#discussion_r562939005, in this PR instead of a followup PR.  @deads2k can you PTAL?

deads2k:
I don't think I'd personally express it this way, but the public API is acceptable.

/approve

pkg/route/apiserver/registry/route/strategy.go
candita: Done

gabemontero:
Hey @candita 

So we recently started seeing errors in our openshift samples testing (stuff that has been around since 3.x of openshift) where creation of simple routes where the host is not specified can fail.  Based on the timing and when this merged, and the nature of this change that I can gleam, I suspect it stems from these changes.

A recent run is at https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/355/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-image-ecosystem/1358801474489946112

Here is the key snippet in the test logs:

```
error: Route.route.openshift.io "rails-postgresql-example" is invalid: spec.host: Invalid value: "rails-postgresql-example-e2e-test-rails-postgresql-repo-test-n22hx.apps.ci-op-x3rbtwji-fe9c0.origin-ci-int-aws.dev.rhcloud.com": host must conform to DNS 1123 naming conventions: [spec.host: Invalid value: "rails-postgresql-example-e2e-test-rails-postgresql-repo-test-n22hx": must be no more than 63 characters]
```

It definitely looks like we are getting the default hostname generation from https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apiserver/simplerouteallocation/plugin.go#L50-L55 where it takes the route name, namespace, and `shard.DNSSuffix`  and concatenates them together.

In this test case, the name is `rails-postgresql-example`, the namespace is `e2e-test-rails-postgresql-repo-test-n22hx`, and the rest must be the `shard.DNSSuffix` from the plugin.... though that sucker alone is almost 63 characters.
But put together they definitely surpass that limit.

My question is (and I did try to see if this was captured in prior PR discussion, but my skimming did not find anything - apologies if you have to regurgitate old discussion) should we not expect our generation logic to be cognizant of the host field restrictions. ... i.e. start stripping characters from the name and/or namespace etc. to get things under 63?

Or is the recommendation that is going to come out from all this is that users are going to have to start explicitly configuring the host field or something of that ilk?

@adambkaplan @bparees FYI

candita:
> ...should we not expect our generation logic to be cognizant of the host field restrictions. ... i.e. start stripping characters from the name and/or namespace etc. to get things under 63?
> Or is the recommendation that is going to come out from all this is that users are going to have to start explicitly configuring the host field or something of that ilk?

Yes, the generation logic will not do the name-shrinking for you. The workaround is to set a custom host name or use the new annotation which allows invalid host names (which make invalid routes).

The solution we decided on was to make the route creation fail faster, without introducing host names that departed from the known formula.  If your test here is creating routes, they would have been invalid routes because the generated host name label is too long, and that would not have been discovered at the route-creation validation, but at the route-admission validation.  

If you require a valid route, you will need to use a custom host name.  If you do not require a valid route, you can use a new route annotation to override the strict validation:  `route.openshift.io/allow-non-dns-compliant-host`, described in https://github.com/openshift/api/blob/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e/route/v1/types.go#L287

gabemontero:
On Mon, Feb 8, 2021 at 6:08 PM Candace Holman <notifications@github.com>
wrote:

> ...should we not expect our generation logic to be cognizant of the host
> field restrictions. ... i.e. start stripping characters from the name
> and/or namespace etc. to get things under 63?
> Or is the recommendation that is going to come out from all this is that
> users are going to have to start explicitly configuring the host field or
> something of that ilk?
>
> Yes, the generation logic will not do the name-shrinking for you. The
> workaround is to set a custom host name or use the new annotation which
> allows invalid host names (which make invalid routes).
>
> The solution we decided on was to make the route creation fail faster,
> without introducing host names that departed from the known formula. If
> your test here is creating routes, they would have been invalid routes
> because the generated host name label is too long, and that would not have
> been discovered at the route-creation validation, but at the
> route-admission validation.
>
> If you require a valid route, you will need to use a custom host name. If
> you do not require a valid route, you can use a new route annotation to
> override the strict validation:
> route.openshift.io/allow-non-dns-compliant-host, described in
> https://github.com/openshift/api/blob/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e/route/v1/types.go#L287
>

Gotcha.  Thanks for the pointers.

PR #173, opened 2021-01-12
Bug 1915018: Handle deleted namespaces in project cache

mfojtik:


mfojtik:
@deads2k PTAL

pkg/project/auth/watch.go
```diff
@@ -147,6 +148,10 @@ func (w *userProjectWatcher) GroupMembershipChanged(namespaceName string, users,
 
 	case hasAccess:
 		namespace, err := w.projectCache.GetNamespace(namespaceName)
+		if kapierrs.IsNotFound(err) {
+			delete(w.knownProjects, namespaceName)
```
deads2k, line 152: This reaction looks wrong.  Is this happening because a namespace was deleted, but we still have access to the deleted namespace?  In such a case, we would expect to send a watch.Deleted event, correct?

Before making that change, you need to be sure that the projectCache is consistent with the namespace lister.

deads2k: Oh, and it should be possible to write an e2e test that will fail this exact way.

pkg/project/cache/cache.go
```diff
@@ -65,9 +66,14 @@ func (p *ProjectCache) GetNamespace(name string) (*corev1.Namespace, error) {
...
 		namespace, err = p.Client.Get(context.TODO(), name, metav1.GetOptions{})
+		// Deleted namespaces can cause error log spam
+		if errors.IsNotFound(err) {
+			klog.V(4).Infof("namespace %s does not exists", name)
+			return nil, err
+		}
 		// the namespace does not exist, so prevent create and update in that namespace
 		if err != nil {
-			return nil, fmt.Errorf("namespace %s does not exist", name)
+			return nil, fmt.Errorf("failed to get namespace %q: %v", name, err)
```
mfojtik, line 76: @deads2k

PR #181, opened 2021-02-01
Bug 1912590: Populate publicDockerImageRepository for new image streams

dmage:
The API server shouldn't modify CacheableObjects as their encoded representation may already be cached. Instead the API server should populate all necessary fields when it creates objects as is already doing for dockerImageRepository.

See also https://github.com/kubernetes/kubernetes/blob/b7fc3f91ea64eff4be673449e08df581eeddf42c/pkg/registry/core/service/storage/storage.go#L128-L143


ricardomaraschini (approved):
/lgtm

PR #192, opened 2021-03-11
limit basic-user cluster role well-known access to oauth-authorization-server only

stlaz:
Upstream started leveraging .well-known endpoints and only allows
serviceaccounts to access them by default. Limit
basic-user access to the `/.well-known/oauth-authorization-server`
endpoint that is needed for authentication into OpenShift.

Counter-part in kube-apiserver: https://github.com/openshift/kubernetes/pull/615

cc @deads2k @sttts


stlaz:
/retest
infra fun

PR #194, opened 2021-03-30
Allow route subdomains to be omitted

smarterclayton:
When user specifies Subdomain without Host, don't default
Host. Follow the same logic as we have for host validation
except we don't allow leniency because that is legacy code.

Mutation follows the same rules as host - must have the
appropriate permissions.

### DIFF
--- a/pkg/route/apis/route/validation/validation.go
+++ b/pkg/route/apis/route/validation/validation.go
@@ -27,18 +27,17 @@ func ValidateRoute(route *routeapi.Route) field.ErrorList {
 }
 
 // validateRoute tests if required fields in the route are set.
-func validateRoute(route *routeapi.Route, check_hostname bool) field.ErrorList {
+func validateRoute(route *routeapi.Route, checkHostname bool) field.ErrorList {
 	//ensure meta is set properly
 	result := validation.ValidateObjectMeta(&route.ObjectMeta, true, ValidateRouteName, field.NewPath("metadata"))
 
 	specPath := field.NewPath("spec")
 
-	if check_hostname && len(route.Spec.Host) > 0 {
+	if checkHostname {
 		if err := validateHost(route); err != nil {
 			result = append(result, err...)
 		}
 	}
-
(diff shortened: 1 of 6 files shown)


pkg/route/apis/route/validation/validation.go
```diff
@@ -93,15 +92,19 @@ func validateRoute(route *routeapi.Route, check_hostname bool) field.ErrorList {
 	return result
 }
 
-// validateHost checks that a route's host name meets DNS requirements.
+// validateHost checks that a route's host name and subdomain meets DNS requirements.
 func validateHost(route *routeapi.Route) field.ErrorList {
-	result := field.ErrorList{}
-	if len(route.Spec.Host) == 0 {
-		return result
+
```
Miciah, line 97: Extra blank line?

pkg/route/apis/route/validation/validation.go
```diff
@@ -93,15 +92,19 @@ func validateRoute(route *routeapi.Route, check_hostname bool) field.ErrorList {
...
+	var result field.ErrorList
+	if len(route.Spec.Host) > 0 {
+		lenient, _ := route.Annotations[routev1.AllowNonDNSCompliantHostAnnotation]
+		result = append(result, library.ValidateHost(route.Spec.Host, lenient, field.NewPath("spec.host"))...)
+	}
+	if len(route.Spec.Subdomain) > 0 {
+		// Subdomain is not lenient because it was never used outside of routes
+		// TODO: fix library-go ValidateHost to take a boolean
+		result = append(result, library.ValidateHost(route.Spec.Host, "", field.NewPath("spec.subdomain"))...)
```
Miciah, line 106:
```suggestion
		result = append(result, library.ValidateHost(route.Spec.Subdomain, "", field.NewPath("spec.subdomain"))...)
```
But `library.ValidateHost(route.Spec.Subdomain, "", ...` isn't right because it will not allow single-label subdomains.

Miciah (commented):
Needs rebase.  I can open a new PR if you don't have time to do it.

PR #197, opened 2021-04-01
[poc] make route admission restrict hsts options

deads2k:
/hold

This is intended to demonstrate possibilities, not to merge.

@candita @Miciah


pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,384 @@
...
+// isRouteHSTSAllowed returns nil if the route is allowed.  Otherwise, returns details and a suggestion in the error
+func isRouteHSTSAllowed(requirements []RequiredHSTS, newRoute *routeapi.Route, namespace *corev1.Namespace) error {
+	for _, requirement := range requirements {
+		if matches, err := requiredHSTSMatchesRoute(requirement, newRoute, namespace); err != nil {
+			return err
+		} else if !matches {
+			continue
+		}
+
+		routeHSTS, err := hstsConfigFromRoute(newRoute)
```
candita, line 100: This doesn't match the RFE, which requests that routes without an HSTS configuration be set with a default for the domain.

candita: See https://issues.redhat.com/browse/NE-310.  We already have a way for users to set HSTS via route annotation.  We need a way for them to set HSTS without individual route annotation.

vendor/github.com/openshift/api/config/v1/types_ingress.go
```diff
@@ -41,13 +43,236 @@ type IngressSpec struct {
...
+	RelatedObjects []ObjectReference `json:"relatedObjects"`
+}
+
+// RequiredRouteAnnotations specifies annotations that are required to be set on
+// newly created routes matching some criteria.
+type RequiredRouteAnnotations struct {
+	// domains is an optional list of domains for which these annotations
+	// are required.  If domains is specified and a route is created with a
+	// spec.host matching one of the domains, the route must specify the
+	// annotations specified in requiredAnnotations.  If domains is empty,
```
candita, line 189: If domains is empty, it probably should ignore RequiredRouteAnnotations entirely.  This will make fewer surprises during migration.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,384 @@
...
+		if err != nil{
+			return err
+		}
+		requirementErr := routeHSTS.meetsRequirements(requirement)
+		if requirementErr != nil {
+			return requirementErr
+		}
+		// if the rule matched, then the API can choose to say that only the first  rule applies.  This will make
+		// allowlisting particular domains possible by enumerating them and saying they have no requirements
```
candita, line 109: Can you please clarify this comment?  It sounds like something good, but I'm not sure.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: We have improved the RFE so that route admission is a must-have.

deads2k:
merged in a better PR by @candita

PR #206, opened 2021-04-29
Bug 1940057: Use listers to find pod info for build logs

otaviof:
Replacing direct usage of Kubernetes client by wiring up Informers for Pods and Builds, and refactoring testing to accommodate those changes.


pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -69,21 +78,25 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
...
 		return nil, errors.NewInvalid(build.Kind("BuildLogOptions"), "", errs)
 	}
-	build, err := r.BuildClient.Builds(apirequest.NamespaceValue(ctx)).Get(ctx, name, metav1.GetOptions{})
+
+	ns := apirequest.NamespaceValue(ctx)
+	build, err := r.BuildLister.Builds(ns).Get(name)
 	if err != nil {
 		return nil, err
 	}
+
```
gabemontero, line 87: need to do a 

`buildCopy := build.DeepCopy`

and then use `buildCopy` instead of `build` for the rest of this method

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -69,21 +78,25 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
...
 	if buildLogOpts.Previous {
 		version := versionForBuild(build)
 		// Use the previous version
 		version--
 		previousBuildName := buildNameForConfigVersion(buildutil.ConfigNameForBuild(build), version)
-		previous, err := r.BuildClient.Builds(apirequest.NamespaceValue(ctx)).Get(ctx, previousBuildName, metav1.GetOptions{})
+		previous, err := r.BuildLister.Builds(ns).Get(previousBuildName)
```
gabemontero, line 93: another deep copy of previous after the err check before it is assigned to build and then used later on

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -121,7 +134,7 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
 
 	// if we can't at least get the build pod, we're not going to get very far, so
 	// error out now.
-	buildPod, err := r.PodClient.Pods(build.Namespace).Get(ctx, buildPodName, metav1.GetOptions{})
+	buildPod, err := r.PodLister.Pods(build.Namespace).Get(buildPodName)
```
gabemontero, line 137: deep copy etc

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -178,7 +191,7 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
 			// assume we are not going to need to iterate again until proven otherwise
 			waitForInitContainers = false
 			// Get the latest version of the pod so we can check init container statuses
-			buildPod, err = r.PodClient.Pods(build.Namespace).Get(ctx, buildPodName, metav1.GetOptions{})
+			buildPod, err = r.PodLister.Pods(build.Namespace).Get(buildPodName)
```
gabemontero, line 194: deep copy etc.

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -268,7 +281,7 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
 			// Wait for the main container to be running, this can take a second after the initcontainers
 			// finish so we have to poll.
 			err := wait.PollImmediate(time.Second, 10*time.Minute, func() (bool, error) {
-				buildPod, err = r.PodClient.Pods(build.Namespace).Get(ctx, buildPodName, metav1.GetOptions{})
+				buildPod, err = r.PodLister.Pods(build.Namespace).Get(buildPodName)
```
gabemontero, line 284: deep copy etc.

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -131,15 +161,22 @@ func TestWaitForBuild(t *testing.T) {
...
 
 	for _, tt := range tests {
+		client := newPodClient()
+		podInformer := fakeCoreV1PodInformer(client)
+
 		build := mockBuild(buildv1.BuildPhasePending, "running", 1)
 		buildClient := buildfakeclient.NewSimpleClientset(build)
+		buildInformer := fakeBuildV1Informer(buildClient)
+
 		fakeWatcher := watch.NewFake()
```
gabemontero, line 171: In conjunction with what @adambkaplan mentioned in slack, and which I think you were on the path for @otaviof if I read the slack exchange correctly, is that I would create a new version for `NewFake()` vs. the default k8s one ... sort of like what was done at https://github.com/openshift/openshift-apiserver/blob/a6220d84588dd916fc13210d3484876832fd37bd/pkg/project/cache/cache.go#L92

then in our "build centric" `fakeWatcher` I would override enough things such that when the client.get is called, you do a DeepCopy like I've mentioned elsewhere in the actual runtime code to avoid cache mutation pain with the SharedInformer stuff in client-go

gabemontero (commented):
So some suggestions @otaviof  ... think they might help if you are still stuck with your current threads of investigation ... though you might have been on this path already (could not 100% tell from the slack exchange)

in the end I think this might be somewhat analogous to the shared informer mutation stuff I recently had to fix in samples operator,  and the demo I will do tomorrow around using `KUBE_CACHE_MUTATION_DETECTOR` set to "true" as an env on the container in question

pkg/build/apiserver/registry/buildlog/rest.go
adambkaplan: Is that necessary if the build object is not changed? If we're only using the build object to find its pod, then I don't see why we need to make a copy.

gabemontero: technically necessary, no

but best practice, one could say it is "safer"

in particular safer to do it now vs. expecting someone making a change in the future that does do a write to remember to go back and do the deep copy now that he will be changing the object

otaviof: Thanks for pointing this out @gabemontero. I have pushed changes to `DeepCopy` objects out of shared cache, so the controller is (thread) safe to interact with them. Please consider.

gabemontero:
yeah @otaviof as was discussed in scrum, all the e2e failures look to me as well like the various, well know e2e flakes and perm fail / often fail errors you can find on sippy

I would say you can remove the WIP whenever you are ready.  The PR title probably could be a tad more descriptive.... "switching to build/v1 informers or some such.

I'll take a pass at the code changes in a bit.

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -40,6 +45,38 @@ func anotherNewPodClient() *fake.Clientset {
 	)
 }
 
+func fakeCoreV1PodInformer(client *fake.Clientset, stopCh <-chan struct{}) informerscorev1.PodInformer {
+	informerFactory := informers.NewSharedInformerFactory(client, 0)
```
gabemontero, line 49: hmm ... are you actually creating real informers for you fake informer?

that could be overkill

did you see this done elsewhere in the apiserver code?

was the fake watcher reference I noted earlier untenable ?

otaviof: I do see other places where we use a fake-client with a regular informer, a couple of examples can be found at [admission_test.go:286](https://github.com/openshift/openshift-apiserver/blob/6a267ebbf4c6d3b3bdacc792c6463ba09f67618f/pkg/image/apiserver/admission/limitrange/admission_test.go#L286-L286) and [imagestreamimport_evaluator_test.go:209](https://github.com/openshift/openshift-apiserver/blob/c8f22885a98f74829c7eb8ea47c2351f77b523a6/pkg/quota/quotaimageexternal/imagestreamtag_evaluator_test.go#L209-L209).

> was the fake watcher reference I noted earlier untenable ?

The previous usage of fake-watcher was causing the racing condition. The informer is based on a watch, and  instantiating a fake-watcher created another. So, we adapted the unit testing to updated the build object directly instead. This subject was [discussed on Slack](https://coreos.slack.com/archives/C014MHHKUSF/p1619692264443400).

pkg/build/apiserver/registry/buildlog/rest_test.go
otaviof: Thanks for the inputs, both the deep-copy and a replacement approach for teh `fakeWatcher` are in place. For the record :-)

otaviof:
/retittle Bug 1940057: Pod and Build Informers

otaviof:
/retitle Bug 1940057: Pod and Build Informers

pkg/build/apiserver/registry/buildlog/rest_test.go
gabemontero: yep I was aware of the original cause of the race and that slack thread

I just thought some tweaks to the hint I provided at https://github.com/openshift/openshift-apiserver/pull/206#discussion_r623368805 would be an alternative to using the "real" client to avoid the race

Perhaps the folk in https://github.com/openshift/openshift-apiserver/blob/master/OWNERS will be amenable to using the real client in the unit test, but at first blush that approach seems atypical to me.  If you have not already @otaviof you might want to reach out to them as well in slack.

otaviof: > I just thought some tweaks to the hint I provided at [#206 (comment)](https://github.com/openshift/openshift-apiserver/pull/206#discussion_r623368805) would be an alternative to using the "real" client to avoid the race

The hints were useful, having a deep-copy in this condition is recommended. However, I need to highlight we have a *fake-client*, with a regular informer. Also, the root of the race-condition is having two watches competing with each other.

xiuwang:
I launched a cluster from this pr, then triggered hundreds of builds. There are no lot of list build messages in the apiserver pod.

pkg/build/apiserver/registry/buildlog/rest_test.go
gabemontero: @otaviof and I talked during scrum.  Let's leave this as is, and see if the apiserver team is good with this and I'm just being too nervous :-)

gabemontero: and I missed the prior examples @otaviof cited ... so yeah we are good here

otaviof:
/assign @sttts

Would you please consider this pull-request? Thanks in advance.

xiuwang:
@otaviof I triggered multiple builds and disconnect apiserver on the cluster launched from pr, then the metric `sum(increase(apiserver_request_total{resource="pods"}[1m])) by (namespace,resource)` only shows default namespace requests, if that's accepted?

http://pastebin.test.redhat.com/962253
image

otaviof:
> @otaviof I triggered multiple builds and disconnect apiserver on the cluster launched from pr, then the metric `sum(increase(apiserver_request_total{resource="pods"}[1m])) by (namespace,resource)` only shows default namespace requests, if that's accepted?
> 
> http://pastebin.test.redhat.com/962253
> image

Thanks for checking this @xiuwang! I tried to reproduce the scenario of running 100 builds, and checking the amount of `LIST` requests. The first image is on a cluster with this pull-request:

image

And, here we have a latest OpenShift 4.7 to compare:

image

So, I think this behavior is expected. We can observe less pressure on the cluster apiserver.

xiuwang:
@otaviof  Thanks for the comparing.
According above result, the openshift-apiserver doesn't increase pressure when a hundred builds are running.

/label qe-approved

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -130,62 +167,124 @@ func TestWaitForBuild(t *testing.T) {
...
 	}
 }
 
-func resourceLocationHelper(buildPhase buildv1.BuildPhase, podPhase string, ctx context.Context, version int) (string, string, string, error) {
+func resourceLocationHelper(ctx context.Context, buildPhase buildv1.BuildPhase, podPhase string, version int) (string, string, string, error) {
+	informersCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	client := newPodClient()
+	podInformer := fakeCoreV1PodInformer(client, informersCtx.Done())
```
adambkaplan, line 277: Assuming using the provided context:
```suggestion
	podInformer := fakeCoreV1PodInformer(client, ctx.Done())
```

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -130,62 +167,124 @@ func TestWaitForBuild(t *testing.T) {
...
+
 	expectedBuild := mockBuild(buildPhase, podPhase, version)
 	buildClient := buildfakeclient.NewSimpleClientset(expectedBuild)
+	buildInformer := fakeBuildV1Informer(buildClient, informersCtx.Done())
```
adambkaplan, line 281: Assuming using the provided context:
```suggestion
	buildInformer := fakeBuildV1Informer(buildClient, ctx.Done())
```

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -199,13 +298,21 @@ func resourceLocationHelper(buildPhase buildv1.BuildPhase, podPhase string, ctx
 	}
 	storage.getSimpleLogsFn = getSimplePodLogsFn
 
+	// artificially waiting for informer's cache synchronization
+	if !cache.WaitForCacheSync(
+		informersCtx.Done(),
```
adambkaplan, line 303: Assuming using provided context:
```suggestion
		ctx.Done(),
```

adambkaplan (commented):
A few questions/nits on the unit tests, main logic looks good.

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -130,62 +167,124 @@ func TestWaitForBuild(t *testing.T) {
...
 	_, err := storage.Get(ctx, build.Name, &buildapi.BuildLogOptions{})
 	if err == nil || !strings.Contains(err.Error(), "timed out") {
 		t.Errorf("Unexpected error result from waitForBuild: %v\n", err)
 	}
 }
 
-func resourceLocationHelper(buildPhase buildv1.BuildPhase, podPhase string, ctx context.Context, version int) (string, string, string, error) {
+func resourceLocationHelper(ctx context.Context, buildPhase buildv1.BuildPhase, podPhase string, version int) (string, string, string, error) {
+	informersCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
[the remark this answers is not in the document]
otaviof: Sure! Those are two different types of context with two different objectives. The original instance is used by the rest-storage to [extract the Kubernetes namespace](https://github.com/openshift/openshift-apiserver/blob/4e3c36dcdf9d8e76998320aba5894b3bb8d16492/pkg/build/apiserver/registry/buildlog/rest.go#L82). And, the `context.WithCancel` was introduced to mitigate the racing-conditions we faced early on, making possible to `cancel()` at the end of each test interaction.

pkg/build/apiserver/registry/buildlog/rest_test.go
otaviof: Please consider [this comment](https://github.com/openshift/openshift-apiserver/pull/206#discussion_r630787588).

pkg/build/apiserver/registry/buildlog/rest_test.go
otaviof: Please consider [this comment](https://github.com/openshift/openshift-apiserver/pull/206#discussion_r630787588).

pkg/build/apiserver/registry/buildlog/rest_test.go
otaviof: Please consider [this comment](https://github.com/openshift/openshift-apiserver/pull/206#discussion_r630787588).

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -130,62 +167,124 @@ func TestWaitForBuild(t *testing.T) {
 		},
 	}
 
+	name := "running"
+	version := 1
+
 	for _, tt := range tests {
```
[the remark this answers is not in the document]
otaviof: Indeed, thanks for the suggestion. Please consider latest changes.

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -130,62 +167,124 @@ func TestWaitForBuild(t *testing.T) {
...
+		}
 		if tt.expectError && err == nil {
 			t.Errorf("%s: Expected an error but got nil from waitFromBuild", tt.name)
 		}
 		if !tt.expectError && err != nil {
-			t.Errorf("%s: Unexpected error from watchBuild: %v", tt.name, err)
+			t.Errorf("%s: Unexpected error from watchBuild: '%v'", tt.name, err)
 		}
+
+		cancel()
```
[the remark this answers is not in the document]
otaviof: Please consider latest changes.

pkg/build/apiserver/registry/buildlog/rest_test.go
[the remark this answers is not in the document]
adambkaplan: makes sense

adambkaplan (approved):
/lgtm

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
```diff
@@ -74,6 +77,10 @@ func NewInformers(kubeInformers kexternalinformers.SharedInformerFactory, kubeCl
 	if err != nil {
 		return nil, err
 	}
+	buildClient, err := buildclient.NewForConfig(kubeClientConfig)
```
deads2k, line 80: this should be a loopback connection.  kubeclients go to kube-apiserver and back.  You can go directly back to localhost.

adambkaplan: @otaviof like this:

```suggestion
	buildClient, err := buildclient.NewForConfig(loopbackClientConfig)
```

adambkaplan (changes requested):

adambkaplan:
@otaviof squash the last two commits and I'll re-tag lgtm.

otaviof:
> @otaviof squash the last two commits and I'll re-tag lgtm.

Thanks for the patch (loopback client), and just pushed the squashed commits.

pkg/build/apiserver/apiserver.go
```diff
@@ -26,11 +27,15 @@ import (
 	"github.com/openshift/openshift-apiserver/pkg/build/apiserver/webhook/generic"
 	"github.com/openshift/openshift-apiserver/pkg/build/apiserver/webhook/github"
 	"github.com/openshift/openshift-apiserver/pkg/build/apiserver/webhook/gitlab"
+	kubeinformers "k8s.io/client-go/informers"
```
adambkaplan, line 30: @otaviof nit - this should be with the other k8s.io imports

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -31,12 +31,16 @@ import (
 	"github.com/openshift/openshift-apiserver/pkg/build/apis/build/validation"
 	"github.com/openshift/openshift-apiserver/pkg/build/apiserver/apiserverbuildutil"
 	buildwait "github.com/openshift/openshift-apiserver/pkg/build/apiserver/registry/wait"
+	corev1listers "k8s.io/client-go/listers/core/v1"
```
adambkaplan, line 34: nit - should be with the other k8s.io imports

pkg/build/apiserver/registry/wait/rest.go
```diff
@@ -30,9 +30,11 @@ type ErrWatchError struct {
...
 // the build ran within timeout, false if it did not, and an error if any other error state occurred.
 // The last observed Build state is returned.
-func WaitForRunningBuild(buildClient buildtypedclient.BuildsGetter, buildNamespace, buildName string, timeout time.Duration) (*buildv1.Build,
-	bool,
-	error) {
+func WaitForRunningBuild(
+	buildClient buildtypedclient.BuildsGetter,
+	buildNamespace, buildName string,
+	timeout time.Duration,
+) (*buildv1.Build, bool, error) {
```
adambkaplan, line 37: It doesn't look like this function did not change its signature, better off leaving it as is.

pkg/build/apiserver/registry/buildlog/rest_test.go
```diff
@@ -10,15 +11,19 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/watch"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
+	"k8s.io/client-go/informers"
+	informerscorev1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes/fake"
-	clientgotesting "k8s.io/client-go/testing"
+
+	"k8s.io/client-go/tools/cache"
```
adambkaplan, line 20: nit - no line break here so that k8s.io imports are grouped together. `go fmt` will sort them alphabetically.

```suggestion
	"k8s.io/client-go/tools/cache"
```

adambkaplan (changes requested):
@otaviof code nits at this point, otherwise looks good.

adambkaplan:
/retitle Bug 1940057: Use listers to find pod info for build logs

pkg/build/apiserver/apiserver.go
otaviof: Done, please consider latest commit. Thanks!

pkg/build/apiserver/registry/wait/rest.go
otaviof: Indeed, moving it back to original.

adambkaplan (approved):
/lgtm

xiuwang:
Launch cluster from pr, then trigger hundreds of builds, check pod LIST request. There is less pressure on cluster apiserver.
image

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -268,12 +284,13 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
...
 			// finish so we have to poll.
 			err := wait.PollImmediate(time.Second, 10*time.Minute, func() (bool, error) {
-				buildPod, err = r.PodClient.Pods(build.Namespace).Get(ctx, buildPodName, metav1.GetOptions{})
+				buildPod, err = r.PodLister.Pods(build.Namespace).Get(buildPodName)
 				if err != nil {
 					s := fmt.Sprintf("error while getting build logs, could not retrieve build pod %s/%s : %v", build.Namespace, buildPodName, err.Error())
 					pipeStreamer.In.Write([]byte(s))
 					return false, err
 				}
+				buildPod = buildPod.DeepCopy()
```
sttts, line 293: Why deep copy? Is it mutated? Where?

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -178,14 +193,15 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
...
-			buildPod, err = r.PodClient.Pods(build.Namespace).Get(ctx, buildPodName, metav1.GetOptions{})
+			buildPod, err = r.PodLister.Pods(build.Namespace).Get(buildPodName)
 			if err != nil {
 				s := fmt.Sprintf("error retrieving build pod %s/%s : %v", build.Namespace, buildPodName, err.Error())
 				// we're sending the error message as the log output so the user at least sees some indication of why
 				// they didn't get the logs they expected.
 				pipeStreamer.In.Write([]byte(s))
 				return
 			}
+			buildPod = buildPod.DeepCopy()
```
sttts, line 204: is it mutated?

sttts: same for all the other deepcopies.

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -69,21 +78,26 @@ func (r *REST) Get(ctx context.Context, name string, opts runtime.Object) (runti
 	if errs := validation.ValidateBuildLogOptions(buildLogOpts); len(errs) > 0 {
 		return nil, errors.NewInvalid(build.Kind("BuildLogOptions"), "", errs)
 	}
-	build, err := r.BuildClient.Builds(apirequest.NamespaceValue(ctx)).Get(ctx, name, metav1.GetOptions{})
+
+	ns := apirequest.NamespaceValue(ctx)
+	build, err := r.BuildLister.Builds(ns).Get(name)
```
sttts, line 83: I can see why we use an informer for pods. I don't see why for builds too.

pkg/build/apiserver/registry/buildlog/rest.go
otaviof: And indeed, I also don't see where it mutates, so let me revert those bits and test.

pkg/build/apiserver/registry/buildlog/rest.go
otaviof: Would it have side effects to have an informer for builds? The reasoning was based on the BZ main theme, reducing the amount of direct apisever requests.

pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go
```diff
@@ -62,6 +63,8 @@ type OpenshiftAPIExtraConfig struct {
 	KubeAPIServerClientConfig *rest.Config
 	KubeInformers             kubeinformers.SharedInformerFactory
 
+	BuildInformers buildinformer.SharedInformerFactory
```
adambkaplan, line 66: We don't need build informers any more

pkg/cmd/openshift-apiserver/openshiftapiserver/config.go
```diff
@@ -228,6 +228,7 @@ func NewOpenshiftAPIConfig(config *openshiftcontrolplanev1.OpenShiftAPIServerCon
 			InformerStart:                      informers.Start,
 			KubeAPIServerClientConfig:          kubeClientConfig,
 			KubeInformers:                      kubeInformers, // TODO remove this and use the one from the genericconfig
+			BuildInformers:                     informers.buildInformers,
```
adambkaplan, line 231: Don't need this any more

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
```diff
@@ -74,6 +77,10 @@ func NewInformers(kubeInformers kexternalinformers.SharedInformerFactory, kubeCl
 	if err != nil {
 		return nil, err
 	}
+	buildClient, err := buildclient.NewForConfig(loopbackClientConfig)
+	if err != nil {
+		return nil, err
+	}
```
adambkaplan, line 83: Not needed any more

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
```diff
@@ -89,6 +96,7 @@ func NewInformers(kubeInformers kexternalinformers.SharedInformerFactory, kubeCl
 		securityInformers:      securityv1informer.NewSharedInformerFactory(securityClient, defaultInformerResyncPeriod),
 		userInformers:          userv1informer.NewSharedInformerFactory(userClient, defaultInformerResyncPeriod),
 		operatorInformers:      operatorinformers.NewSharedInformerFactory(operatorClient, defaultInformerResyncPeriod),
+		buildInformers:         buildinformer.NewSharedInformerFactory(buildClient, defaultInformerResyncPeriod),
```
adambkaplan, line 99: Not needed any more

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
```diff
@@ -137,4 +145,5 @@ func (i *InformerHolder) Start(stopCh <-chan struct{}) {
 	i.securityInformers.Start(stopCh)
 	i.userInformers.Start(stopCh)
 	i.operatorInformers.Start(stopCh)
+	i.buildInformers.Start(stopCh)
```
adambkaplan, line 148: Not needed any more

pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go
```diff
@@ -96,6 +99,9 @@ func (c *OpenshiftAPIExtraConfig) Validate() error {
 	if c.KubeInformers == nil {
 		ret = append(ret, fmt.Errorf("KubeInformers is required"))
 	}
+	if c.BuildInformers == nil {
+		ret = append(ret, fmt.Errorf("BuildInformers is required"))
+	}
```
adambkaplan, line 104: Not needed - can be removed

pkg/build/apiserver/registry/buildlog/rest.go
```diff
@@ -48,11 +50,11 @@ const defaultTimeout time.Duration = 30 * time.Second
...
 // URL to which the request shall be redirected in order to get build logs.
-func NewREST(buildClient buildtypedclient.BuildsGetter, podClient kubetypedclient.PodsGetter) *REST {
+func NewREST(buildClient buildtypedclient.BuildsGetter, podClient kubetypedclient.PodsGetter, podLister corev1listers.PodLister) *REST {
 	r := &REST{
-		BuildClient: buildClient,
-		PodClient:   podClient,
-		Timeout:     defaultTimeout,
+		PodClient: podClient,
+		PodLister: podLister,
+		Timeout:   defaultTimeout,
```
adambkaplan, line 57: Missed setting the BuildClient here - this is probably why all build tests are bombing out.

```suggestion
		BuildClient: buildClient,
		PodClient:   podClient,
		PodLister:   podLister,
		Timeout:     defaultTimeout,
```

adambkaplan (changes requested):
Looks like a few things were missed when the build informers were ripped out.

adambkaplan:
/retest

infra flake installing the cluster

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
```diff
@@ -38,6 +39,7 @@ type InformerHolder struct {
 	securityInformers      securityv1informer.SharedInformerFactory
 	userInformers          userv1informer.SharedInformerFactory
 	operatorInformers      operatorinformers.SharedInformerFactory
+	buildInformers         buildinformer.SharedInformerFactory
```
adambkaplan, line 42: buildInformers is still here - not needed

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
adambkaplan: Still here - needs to be removed

adambkaplan (changes requested):
Still have a few dangling references to BuildInformers

pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go
otaviof: Thanks!

adambkaplan (approved):
/lgtm

cuppett:
@deads2k @sttts @mfojtik can we get a review/approval here before code freeze this week?

pkg/build/apiserver/apiserver.go
```diff
@@ -31,6 +32,8 @@ import (
 type ExtraConfig struct {
 	KubeAPIServerClientConfig *restclient.Config
 
+	KubeInformers kubeinformers.SharedInformerFactory
```
sttts, line 35: RecommendedConfig has this already

sttts:
Please use RecommendedConfig's informers. Otherwise sgtm.

pkg/build/apiserver/apiserver.go
otaviof: Thanks! Removed this entry and using from existing instance. Please consider last push.

adambkaplan (approved):
/lgtm

bparees:
/override ci/prow/e2e-aws-upgrade

(this upgrade failure is not related to this pr)

PR #207, opened 2021-05-05
Bug 1957261: BUILD-186 godoc cleanup

gabemontero:
@adambkaplan and @rolfedh both had godoc changes for the new api that were submitted just as the openshift/api PR was merging

This PR pulls in the subsequent openshift/api godoc changes that just merged into openshift-api server so that the end result of `oc explain` displaying the preferred godoc for the new fields occurs.

NOTE: the `make update` invocation following the full openshift/api bump did result in as clean and minimal of a change as desired.  `hack/openapi-violation.list` in particular was updated.  Meaning the non-build changes between the last openshift/api level and the commit with the godoc changes may not be totally clean.

We'll see how the various e2e's shake out and go from there.


gabemontero:
> NOTE: the `make update` invocation following the full openshift/api bump did result in as clean and minimal of a change as desired. `hack/openapi-violation.list` in particular was updated. Meaning the non-build changes between the last openshift/api level and the commit with the godoc changes may not be totally clean.
> 
> We'll see how the various e2e's shake out and go from there.

OK the verify job is happy at least

gabemontero:
unrelated flakes

/retest

adambkaplan:
`error: default cluster has no server defined` - is this a known issue/flake?

adambkaplan (approved):
/lgtm

gabemontero:
> `error: default cluster has no server defined` - is this a known issue/flake?

that last run and that error are different that the prior failures where the tests at least ran but hit issues in non-build components nor components touched by `make update` after the openshift/api bump

if it occurs with some frequency I'll go down the sippy path

also, e2e-cmd it note required and so can be skipped if we are all confident errors are not related to this change

/retest

also need to see who from https://github.com/openshift/openshift-apiserver/blob/master/OWNERS wants to look at and approve this

gabemontero:
and e2e-cmd is green now

PR #208, opened 2021-05-11
BUILD-87: Add Build Volumes API with validation

coreydaley:
https://issues.redhat.com/browse/BUILD-87


pkg/build/apis/build/v1/zz_generated.conversion.go
```diff
@@ -1967,6 +1967,8 @@ func autoConvert_v1_DockerBuildStrategy_To_build_DockerBuildStrategy(in *v1.Dock
 		out.BuildArgs = nil
 	}
 	out.ImageOptimizationPolicy = (*build.ImageOptimizationPolicy)(unsafe.Pointer(in.ImageOptimizationPolicy))
+	// WARNING: in.Volumes requires manual conversion: does not exist in peer-type
+	// WARNING: in.VolumeMounts requires manual conversion: does not exist in peer-type
```
adambkaplan, line 1971: @coreydaley You need to add a manual conversion - odd that it is not being auto-generated.

adambkaplan (changes requested):

pkg/build/apis/build/v1/zz_generated.conversion.go
coreydaley: This should be fixed in the current push

pkg/build/apis/build/types.go
```diff
@@ -1383,3 +1399,65 @@ type SecretSpec struct {
...
+	BuildVolumeSourceTypeConfigMap BuildVolumeSourceType = "ConfigMap"
+)
+
+// BuildVolumeSource represents the source of a volume to mount
+// Only one of its supported types may be specified at any given time.
+type BuildVolumeSource struct {
+
+	// type is the BuildVolumeSourceType for the volume source.
+	// Type must match the populated volume source, and if not specified will be inferred on admission.
+	Type BuildVolumeSourceType
```
adambkaplan, line 1435: One thing I noticed with the build strategy types - they exist in the kube API, but do _not_ exist in the internal representation. We use a conversion function to populate this type discriminator in `conversion.go`.

See https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/apis/build/v1/conversion.go#L137-L154

gabemontero:
Did you run `make update` @coreydaley and store those changes in a separate commit?  It was hard for me to tell with what you currently have here.

If you have not, you'll need that to get your `ci/prow/verify` runs clean.

adambkaplan:
/hold

Given the complexity of the feature and the time remaining before 4.8 reaches "final freeze" period, we are going to hold this until master opens for 4.9

pkg/build/apis/build/v1/zz_generated.conversion.go
```diff
@@ -1697,6 +1727,112 @@ func Convert_build_BuildTriggerPolicy_To_v1_BuildTriggerPolicy(in *build.BuildTr
...
+}
+
+// Convert_build_BuildVolumeMount_To_v1_BuildVolumeMount is an autogenerated conversion function.
+func Convert_build_BuildVolumeMount_To_v1_BuildVolumeMount(in *build.BuildVolumeMount, out *v1.BuildVolumeMount, s conversion.Scope) error {
+	return autoConvert_build_BuildVolumeMount_To_v1_BuildVolumeMount(in, out, s)
+}
+
+func autoConvert_v1_BuildVolumeSource_To_build_BuildVolumeSource(in *v1.BuildVolumeSource, out *build.BuildVolumeSource, s conversion.Scope) error {
+	// INFO: in.Type opted out of conversion generation
+	if in.Secret != nil {
```
adambkaplan, line 1782: @coreydaley I think this is why the generation nit test is failing. Per my prior [comment](https://github.com/openshift/openshift-apiserver/pull/208/files#r632500215), I think what you need is the following:

1. Remove `Type` from the internal representation for BuildVolumeMount
2. Re-run the generated conversions -  this _should_ generate the autoConvert functions
3. Create a separate `Convert_build_BuildVolumeMount_To_v1_BuildVolumeMount` in `conversions.go` which populates the Type field.

adambkaplan (changes requested):

pkg/build/apis/build/v1/zz_generated.conversion.go
coreydaley: So, if I remove the `Type` field, then how would I populate it?
Also, do you mean `BuildVolumeSource`?

adambkaplan:
@coreydaley looks like the upgrade tests are failing for unrelated reasons. I think it is time to start getting QE to test this new feature, and to write the CI tests that we will add to origin.

@xiuwang for testing this feature, we will want to modify the existing build tests for Secret and ConfigMap build inputs to use the new volumes API. BuildConfig specs will look like this:

```yaml
spec:
  source:
    git: ...
  strategy:
    sourceStrategy:
      ...
      volumes:
        - name: some-secret
          source:
            secret:
              secretName: my-secret
          mounts:
             - destinationPath: /var/run/secret/some-secret
               readOnly: true 
```

PRs needed to test this feature:
- openshift/openshift-apiserver#208
- openshift/openshift-controller-manager#183
- openshift/builder#245

adambkaplan (commented):
/approve

Marking approve to indicate we are ready for QE to begin their verification.

adambkaplan:
Given the first test by @xiuwang, we need to add validation on the volume name. It needs to

Validations should be added here (using a common method):

1. https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/apis/build/validation/validation.go#L563-L569
2. https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/apis/build/validation/validation.go#L519-L547

Upstream k8s/apimachinery has a method which lets us validate the name is an appropriate DNS label (required by kube for volumes): https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L187

coreydaley:
@adambkaplan Added validation and tests in current push. ptal

xiuwang:
I have test the feature following test cases https://url.corp.redhat.com/buildvolumes, no critial issue found.

/label qe-approved

adambkaplan:
@coreydaley based on the error message from @xiuwang in JIRA, perhaps this PR needs to be rebased?

coreydaley:
@adambkaplan @gabemontero ptal,

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+		}
+		if strings.Contains(vm.DestinationPath, ":") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "destination path must not contain ':'"))
+		}
+		if !filepath.IsAbs(vm.DestinationPath) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "must be an absolute path"))
+		}
+		if strings.HasPrefix(path.Clean(vm.DestinationPath), "..") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "destination path must not start with '..'"))
+		}
```
gabemontero, line 843: the godoc mentions for the destination path `can not collide with a destination path generated by the builder process`

I'm assuming there is no way for the apiserver to anticipate a path generated down the line by the builder process.  But just in case I'm mentioning it here @coreydaley in case you are aware of perhaps a hard coded path generated by the builder process that the apiserver should check for.

gabemontero (commented):
only one question which may be a no-op @coreydaley .... let me know what your think

pkg/build/apis/build/validation/validation.go
coreydaley: This code only validates that everything has been submitted to the apiserver correctly. If there is a collision, the build will have to fail and report it's status. There are some paths that are used, but it would be a nightmare to keep them in sync here.

gabemontero: figured it was along those lines ... thanks for confirming @coreydaley

gabemontero: I'll add the /lgtm .... if @adambkaplan finds something I did not cite, I'll let him cancel the lgtm

gabemontero:
I'll add the 

/lgtm

and let @adambkaplan cancel it if he finds something I missed @coredaley

of course you'll need to reach out to apiserver team or maybe @bparees for approve

adambkaplan (approved):
/lgtm

adambkaplan:
/retitle BUILD-87: Add Build Volumes API with validation

hack/openapi-violation.list
```diff
@@ -48,17 +48,20 @@ API rule violation: list_type_missing,github.com/openshift/api/build/v1,BuildSou
 API rule violation: list_type_missing,github.com/openshift/api/build/v1,BuildSpec,TriggeredBy
 API rule violation: list_type_missing,github.com/openshift/api/build/v1,BuildStatus,Conditions
 API rule violation: list_type_missing,github.com/openshift/api/build/v1,BuildStatus,Stages
+API rule violation: list_type_missing,github.com/openshift/api/build/v1,BuildVolume,Mounts
```
sttts, line 51: can we avoid these?

vendor/github.com/openshift/api/build/v1/types.go
```diff
@@ -773,6 +780,12 @@ type SourceBuildStrategy struct {
 	// deprecated json field, do not reuse: runtimeArtifacts
 	// +k8s:protobuf-deprecated=runtimeArtifacts,8
 
+	// volumes is a list of input volumes that can be mounted into the builds runtime environment.
+	// Only a subset of Kubernetes Volume sources are supported by builds.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes
+	// +patchMergeKey=name
+	// +patchStrategy=merge
+	Volumes []BuildVolume `json:"volumes,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,9,opt,name=volumes"`
```
sttts, line 788: which kind of list is this? You have to declare the liist type.

vendor/github.com/openshift/api/build/v1/types.go
```diff
@@ -742,6 +742,13 @@ type DockerBuildStrategy struct {
 	// policy. An additional experimental policy 'SkipLayersAndWarn' is the same as
 	// 'SkipLayers' but simply warns if compatibility cannot be preserved.
 	ImageOptimizationPolicy *ImageOptimizationPolicy `json:"imageOptimizationPolicy,omitempty" protobuf:"bytes,8,opt,name=imageOptimizationPolicy,casttype=ImageOptimizationPolicy"`
+
+	// volumes is a list of input volumes that can be mounted into the builds runtime environment.
+	// Only a subset of Kubernetes Volume sources are supported by builds.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes
+	// +patchMergeKey=name
+	// +patchStrategy=merge
+	Volumes []BuildVolume `json:"volumes,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,9,opt,name=volumes"`
```
sttts, line 751: which kind of list is this? You have to declare the liist type.

vendor/github.com/openshift/api/build/v1/types.go
```diff
@@ -1335,3 +1348,62 @@ type SecretSpec struct {
...
+// BuildVolume describes a volume that is made available to build pods,
+// such that it can be mounted into buildah's runtime environment.
+// Only a subset of Kubernetes Volume sources are supported.
+type BuildVolume struct {
+	// name is a unique identifier for this BuildVolume.
+	// It must conform to the Kubernetes DNS label standard and be unique within the pod.
+	// Names that collide with those added by the build controller will result in a
+	// failed build with an error message detailing which name caused the error.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
```
sttts, line 1361: is this required? Needs a tag.

vendor/github.com/openshift/api/build/v1/types.go
```diff
@@ -1335,3 +1348,62 @@ type SecretSpec struct {
...
+
+	// source represents the location and type of the mounted volume.
+	Source BuildVolumeSource `json:"source" protobuf:"bytes,2,opt,name=source"`
+
+	// mounts represents the location of the volume in the image build container
+	Mounts []BuildVolumeMount `json:"mounts" protobuf:"bytes,3,opt,name=mounts"`
```
sttts, line 1367: Which kind of list is this? Specify list-type.

vendor/github.com/openshift/api/build/v1/types.go
```diff
@@ -1335,3 +1348,62 @@ type SecretSpec struct {
...
+type BuildVolume struct {
+	// name is a unique identifier for this BuildVolume.
+	// It must conform to the Kubernetes DNS label standard and be unique within the pod.
+	// Names that collide with those added by the build controller will result in a
+	// failed build with an error message detailing which name caused the error.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+
+	// source represents the location and type of the mounted volume.
+	Source BuildVolumeSource `json:"source" protobuf:"bytes,2,opt,name=source"`
```
sttts, line 1364: is this required?

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
 	return patch, nil
 }
+
+func validateBuildVolume(volume buildapi.BuildVolume, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(volume.Name) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), ""))
+	}
+	if errs := kvalidation.IsDNS1123Label(volume.Name); len(errs) != 0 {
```
sttts, line 803: else ...

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+		foundVolumeSources = append(foundVolumeSources, string(buildapi.BuildVolumeSourceTypeSecret))
+	}
+	if volume.Source.ConfigMap != nil {
+		foundVolumeSources = append(foundVolumeSources, string(buildapi.BuildVolumeSourceTypeConfigMap))
+	}
+	if len(foundVolumeSources) == 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source"), volume.Source, "must specify one volume source"))
+	}
+	if len(foundVolumeSources) > 1 {
+		allErrs = append(allErrs, field.TooMany(fldPath.Child("source"), len(foundVolumeSources), 1))
```
sttts, line 823: I don't think TooMany fits here. Compare:
```
// TooMany returns a *Error indicating "too many". This is used to
// report that a given list has too many items. This is similar to TooLong,
// but the returned error indicates quantity instead of length.
```

This is not a list, but a union.

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+	}
+	if len(foundVolumeSources) != 0 && foundVolumeSources[0] != string(volume.Source.Type) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("type"), volume.Source, "source Type and specified type must match"))
```
sttts, line 826: lower-case "Type". This is not about the Golang types, but about json.

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+	}
+	if len(volume.Mounts) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("mounts"), "must supply at least one mount"))
+	}
+	for i, vm := range volume.Mounts {
+		if len(vm.DestinationPath) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("destinationPath"), "must not be blank"))
```
sttts, line 833: field path is wrong

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+		if strings.Contains(vm.DestinationPath, ":") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "destination path must not contain ':'"))
+		}
+		if !filepath.IsAbs(vm.DestinationPath) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "must be an absolute path"))
+		}
+		if strings.HasPrefix(path.Clean(vm.DestinationPath), "..") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "destination path must not start with '..'"))
+		}
+	}
```
sttts, line 844: would have expected some regex based validation first that defines the string sets positively rather than excluding a few corner cases.

vendor/github.com/openshift/api/build/v1/types.go
coreydaley: updated in next push

hack/openapi-violation.list
coreydaley: fixed in next push

vendor/github.com/openshift/api/build/v1/types.go
coreydaley: addressed via https://github.com/openshift/api/pull/954

vendor/github.com/openshift/api/build/v1/types.go
coreydaley: addressed via https://github.com/openshift/api/pull/954

vendor/github.com/openshift/api/build/v1/types.go
coreydaley: addressed via https://github.com/openshift/api/pull/954

vendor/github.com/openshift/api/build/v1/types.go
coreydaley: addressed via https://github.com/openshift/api/pull/954

pkg/build/apis/build/validation/validation.go
coreydaley: Could you elaborate on why this needs an `else` branch?

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

pkg/build/apis/build/validation/validation.go
coreydaley: This pattern was copied from https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/apis/build/validation/validation.go#L367-L375

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

vendor/github.com/openshift/api/build/v1/types.go
coreydaley: addressed via openshift/api#954

pkg/build/apis/build/validation/validation.go
sttts: so we say: path can be any utf8 string minus those we check. Does e.g. Windows support matter here?

pkg/build/apis/build/validation/validation.go
sttts: is empty string a DNS1123label? You will get double error now with your code.

coreydaley: doh! I violated my "always read the next line" rule.
Updated in next push

RickJWagner:
PLM agrees.

pkg/build/apis/build/validation/validation.go
adambkaplan: Windows support does not matter yet. We only support builds for Linux container images today.

adambkaplan: (perhaps we need a TODO on that validation if we ever decide to support Windows containers for builds).

coreydaley: @sttts Since windows support is not an issue, can this pull request move forward?

sttts: yes. Just note that you will never be ableto strengthen the validation here. It will stay as open as it is now.

coreydaley: @adambkaplan Are you ok with ^ ?

adambkaplan: I think this is fine.

adambkaplan (approved):
/lgtm

coreydaley:
@sttts I think we are ready for an approval, the e2e-aws-upgrade test failures are unrelated and being discussed on slack.

pkg/build/apis/build/validation/validation.go
sttts: What I had meant original is still not the case:

```
if len(volume.Name) == 0 {
  allErrs = append(allErrs, field.Required(fldPath.Child("name"), ""))
} else if errs := kvalidation.IsDNS1123Label(volume.Name); len(errs) != 0 {
   ...
}
```

Otherwise you get `["volume.Name is required", "volume.Name is not valid DNS 1123 label"]`.

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,54 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+	allErrs := field.ErrorList{}
+
+	if len(volume.Name) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), ""))
+	}
+	if errs := kvalidation.IsDNS1123Label(volume.Name); len(errs) != 0 {
+		for _, e := range errs {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), volume, e))
+		}
+	} else if len(volume.Source.Type) == 0 {
```
sttts, line 807: why this `else`?

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,54 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+		foundVolumeSources = append(foundVolumeSources, string(buildapi.BuildVolumeSourceTypeConfigMap))
+	}
+	if len(foundVolumeSources) == 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source"), volume.Source, "must specify one volume source"))
+	}
+	if len(foundVolumeSources) > 1 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source"), volume.Source, "only one volume source is allowed"))
+	}
+	if len(foundVolumeSources) != 0 && foundVolumeSources[0] != string(volume.Source.Type) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("type"), volume.Source, "source type and specified type must match"))
```
sttts, line 825: `volume.Source` is not the value at the field path.

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,54 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+
+func validateBuildVolume(volume buildapi.BuildVolume, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(volume.Name) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), ""))
+	}
+	if errs := kvalidation.IsDNS1123Label(volume.Name); len(errs) != 0 {
+		for _, e := range errs {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), volume, e))
```
sttts, line 805: volume is not the value at the field path

pkg/build/apis/build/validation/validation.go
coreydaley: updated in current push

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,54 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+		allErrs = append(allErrs, field.Required(fldPath.Child("mounts"), "must supply at least one mount"))
+	}
+	for i, vm := range volume.Mounts {
+		if len(vm.DestinationPath) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("mounts").Child("destinationPath"), "must not be blank"))
+		}
+		if strings.Contains(vm.DestinationPath, ":") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "destination path must not contain ':'"))
+		}
+		if !filepath.IsAbs(vm.DestinationPath) {
```
sttts, line 837: I assume `IsAbs` is false for empty string. You will get two errors now for empty string.

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

sttts:
A couple of validation formalities are wrong.

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

coreydaley:
@sttts All comments have been addressed, thanks for the review

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+	}
+	if len(volume.Mounts) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("mounts"), "must supply at least one mount"))
+	}
+	for i, vm := range volume.Mounts {
+		if len(vm.DestinationPath) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("mounts").Child("destinationPath"), "must not be blank"))
+		} else {
+			if strings.Contains(vm.DestinationPath, ":") {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("mounts").Index(i).Child("destinationPath"), vm.DestinationPath, "destination path must not contain ':'"))
```
sttts, line 835: cosmetics: all these strings must be written wir the field path as subjects. I.e. here: `must not contain ":"`. Otherwise you will get `mounts.destinationPath "bla:bla" destination path must not contain ':' `

sttts: There are some more like that above and below

coreydaley: updated in current push

pkg/build/apis/build/validation/validation.go
```diff
@@ -787,3 +793,55 @@ func CreateBuildPatch(older, newer *buildapi.Build) ([]byte, error) {
...
+		foundVolumeSources = append(foundVolumeSources, string(buildapi.BuildVolumeSourceTypeConfigMap))
+	}
+	if len(foundVolumeSources) == 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source"), volume.Source, "must specify one volume source"))
+	}
+	if len(foundVolumeSources) > 1 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source"), volume.Source, "only one volume source is allowed"))
+	}
+	if len(foundVolumeSources) != 0 && foundVolumeSources[0] != string(volume.Source.Type) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("type"), volume.Source.Type, "source type and specified type must match"))
```
sttts, line 825: this and the one above is still not of shape `must ....`. Follow-up is fine.

PR #213, opened 2021-06-03
Bug 1956826: add restriction note to docker strategy build args

gabemontero:
/assign @adambkaplan


jitendar-singh:
jsingh@fugaku  ~/Downloads/openshift-install-linux  oc explain build.spec.strategy.dockerStrategy.buildArgs
KIND:     Build
VERSION:  build.openshift.io/v1
RESOURCE: buildArgs <[]Object>
DESCRIPTION:
     buildArgs contains build arguments that will be resolved in the Dockerfile.
     See https://docs.docker.com/engine/reference/builder/#/arg for more
     details. NOTE: Only the 'name' and 'value' fields are supported. Any
     settings on the 'valueFrom' field are ignored.
     EnvVar represents an environment variable present in a Container.
FIELDS:
   name	<string> -required-
     Name of the environment variable. Must be a C_IDENTIFIER.
   value	<string>
     Variable references $(VAR_NAME) are expanded using the previous defined
     environment variables in the container and any service environment
     variables. If a variable cannot be resolved, the reference in the input
     string will be unchanged. The $(VAR_NAME) syntax can be escaped with a
     double $$, ie: $$(VAR_NAME). Escaped references will never be expanded,
     regardless of whether the variable exists or not. Defaults to "".
   valueFrom	<Object>
     Source for the environment variable's value. Cannot be used if value is not
     empty.

gabemontero:
failed to acquire aws quota lease e2e-aws-builds

/test e2e-aws-builds

PR #217, opened 2021-06-16
Bug 1972383: openshift authorization proxy: escape header key values

stlaz:
The authorization.openshift.io RBAC proxy is taking values from
the Extra field of a UserInfo of a user that made a request against
this API. The fields of the map may generally contain values that
cannot appear in an HTTP header, like '/' that's commonly separating
an API annotation from the API version or resource.

Copy the k8s.io header escaping function and use it in the proxy
that's handling the OpenShift authorization API.

/cc @s-urbaniak @sttts 

/cherry-pick release-4.7

### DIFF
--- a/pkg/client/impersonatingclient/impersonate.go
+++ b/pkg/client/impersonatingclient/impersonate.go
@@ -1,7 +1,9 @@
 package impersonatingclient
 
 import (
+	"fmt"
 	"net/http"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/client/impersonatingclient/impersonate.go
```diff
@@ -86,3 +88,110 @@ func (c *impersonatingRESTClient) APIVersion() schema.GroupVersion {
...
+	return int(b) < len(legalHeaderKeyBytes) && legalHeaderKeyBytes[b]
+}
+
+func shouldEscape(b byte) bool {
+	// url.PathUnescape() returns an error if any '%' is not followed by two
+	// hexadecimal digits, so we'll intentionally encode it.
+	return !legalHeaderByte(b) || b == '%'
+}
+
+func headerKeyEscape(key string) string {
```
s-urbaniak, line 102: can we add a reference to client-go?

stlaz: added

osherdp:
/lgtm
works on my env

PR #222, opened 2021-06-23
Bug 1974651: Remove client for Docker v1 API

dmage:
The Docker Registry v1 API has been deprecated for a while and should not be used anymore.


ricardomaraschini (approved):
/lgtm

PR #223, opened 2021-06-24
WIP: BUILD-159:fake openshift/api bump

gabemontero:
Before https://github.com/openshift/api/pull/952 merges, testing out these changes via clusterbot where in theory the k8s apimachinery validation producing the error:

```
The ClusterCSIDriver "csi.projected.shared.resources.openshift.io" is invalid: metadata.name: Unsupported value: "csi-driver-projected-resource.openshift.io": supported values: "ebs.csi.aws.com", "disk.csi.azure.com", "pd.csi.storage.gke.io", "cinder.csi.openstack.org", "csi.vsphere.vmware.com", "manila.csi.openstack.org", "csi.ovirt.org", "csi.kubevirt.io"
```

even with both CVO and CSO disabled per https://github.com/openshift/gcp-pd-csi-driver-operator#quick-start so that I can try running our new csi driver operator locally.


gabemontero:
duh, we are talking CRDs, not new openshift aggregated API types.

I've hacked the wrong apiserver

PR #224, opened 2021-06-29
Bug 1986408: NE-310 HSTS Route Admission Plugin

candita:
Administrators who are tasked with route management and/or regulatory compliance face a number of issues with regard to enforcing HSTS.  For efficiency and protection against configuration errors, they have requested to automatically and globally enable HSTS on the basis of the cluster `Ingress` domains. However, because enabling HSTS automatically and globally can cause disruption of service on a wide scale, they should also be able to audit and change HSTS configuration without further outages.  Finally, they should be able to consistently apply the same HSTS configuration and predict the outcome on any cluster.

- Register a new plugin route.openshift.io/RequiredRouteAnnotations 
- Add a new admissions package called requiredrouteannotations, with methods to validate route annotation configuration
- Add unit tests

Enhancement proposal: https://github.com/openshift/enhancements/pull/749


pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,337 @@
...
+	if o.nsLister == nil {
+		return fmt.Errorf(pluginName + " plugin needs a namespace lister")
+	}
+	if len(o.cachesToSync) < 3 {
+		return fmt.Errorf(pluginName + " plugin missing informer synced functions")
+	}
+	return nil
+}
+
+func NewRequiredRouteAnnotations() (admission.Interface, error) {
```
sanchezl, line 152:
```suggestion
func NewRequiredRouteAnnotations() (*requiredRouteAnnotations, error) {
```
It's better to return the concrete type here.

candita:
Running step e2e-aws-ipi-install-install.    
{"component":"entrypoint","file":"prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2021-07-21T23:34:21Z"}

/test e2e-aws

candita:
info: Loading sha256:a4ae2a1ae18b0b4ecf96994e04aa8770d72a611b3cd73460ebb91fde23841910 vsphere-problem-detector
error: unable to parse image registry.build01.ci.openshift.org/ci-op-ml44p4ql/stable-initial@sha256:cceb4cbb8e255ea598f368cd7ed2a09b9b85a5abc86df9dd16a0364837c42662: cannot retrieve image configuration for manifest sha256:cceb4cbb8e255ea598f368cd7ed2a09b9b85a5abc86df9dd16a0364837c42662: received unexpected HTTP status: 500 Internal Server Error

/test e2e-aws-upgrade

hack/update-generated-openapi.sh
```diff
@@ -25,7 +25,7 @@ KUBE_INPUT_DIRS=(
 )
 ORIGIN_INPUT_DIRS=(
   $(
-    grep --color=never -rl '+k8s:openapi-gen=' vendor/github.com/openshift/api | \
+    grep --color=never -rl '+k8s:openapi-gen=true' vendor/github.com/openshift/api | \
```
Miciah, line 28: Were the changes in `hack/update-generated-openapi.sh` required after all?

candita: They are not required by my code, but this could be an issue in the future.  These two files show up as input files and should not, because `+k8s:openapi-gen=false` there:
k8s.io/api/admission/v1
k8s.io/api/admission/v1beta1

candita: If you think it's too risky, I can add to a separate PR

Miciah: I don't object to keeping this, but I cannot speak for other reviewers.

candita: I'll put it in a different PR asap

candita:
Managed cluster should have same number of Machines and Nodes
Many 137 error codes, will retry
/test e2e-aws

Miciah:
[e2e-cmd](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_openshift-apiserver/224/pull-ci-openshift-openshift-apiserver-master-e2e-cmd/1418657982308159488) failed on `[sig-cli][Feature:LegacyCommandTests][Disruptive][Serial] test-cmd: test/cmd/basicresources.sh` with the following:

    The Service "frontend" is invalid: spec.clusterIPs: Invalid value: []string{"172.30.182.32"}: failed to allocated ip:172.30.182.32 with error:provided IP is already allocated

/test e2e-cmd

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+	matchesDom := matchesDomain(requirement.DomainPatterns, routeDomains)
+
+	return matchesNamespace, matchesDom, nil
+}
+
+// Check all of the required domainMatcher patterns against all provided domains,
+// first match returns true.  If none match, return false.
+func matchesDomain(domainMatchers []string, domains []string) bool {
+	for _, matcher := range domainMatchers {
+		match := regexp.MustCompile(fmt.Sprintf("^%s$", matcher))
```
sanchezl, line 340: The documentation for `RequiredHSTSPolicy.DomainPatterns` describes the pattern syntax as a simplified glob-like pattern, not a regex.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+var _ = initializer.WantsExternalKubeInformerFactory(&requiredRouteAnnotations{})
+var _ = admission.ValidationInterface(&requiredRouteAnnotations{})
+var _ = openshiftapiserveradmission.WantsOpenShiftConfigInformers(&requiredRouteAnnotations{})
+var _ = openshiftapiserveradmission.WantsOpenShiftRouteInformers(&requiredRouteAnnotations{})
+
+const hstsAnnotation = "haproxy.router.openshift.io/hsts_header"
+
+// Validate ensures that routes specify required annotations, and returns nil if valid.
+func (o *requiredRouteAnnotations) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) (err error) {
+	if a.GetResource().GroupResource() != routev1.Resource("routes") {
```
sanchezl, line 67:
```suggestion
	if a.GetResource().GroupResource() != route.Resource("routes") {
```

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil
+	}
+	newobject, isRoute := a.GetObject().(*routeapi.Route)
```
sanchezl, line 70:
```suggestion
	newObject, isRoute := a.GetObject().(*routeapi.Route)
```

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+		if c.includeSubDomains {
+			return fmt.Errorf("includeSubDomains must not be specified")
+		}
+	}
+
+	return nil
+}
+
+// Check if the route matches the required domain/namespace in the HSTS Policy
+func requiredNamespaceDomainMatchesRoute(requirement configv1.RequiredHSTSPolicy, route *routeapi.Route, namespace *corev1.Namespace) (bool, bool, error) {
```
sanchezl, line 321: Consider just returning a single `bool` for a match/no match. You don't seem to have a need to distinguish the values outside this method.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+	if o.nsLister == nil {
+		return fmt.Errorf(pluginName + " plugin needs a namespace lister")
+	}
+	if len(o.cachesToSync) < 3 {
+		return fmt.Errorf(pluginName + " plugin missing informer synced functions")
+	}
+	return nil
+}
+
+func NewRequiredRouteAnnotations() (*requiredRouteAnnotations, error) {
```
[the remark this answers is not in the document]
candita: This is a `Factory` and requires that unused parameter for the `Plugins.Registry`

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: I hadn't intended for that to happen.  I could correct the documentation later, if that's acceptable.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: Do you want me to change the import alias from `routev1` to `route` here?  And for all the other places I've used the v1 suffix?

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: done, thanks!

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: done, thanks

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: Done, thanks!  Thanks @frobware too.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	klog.Infof("allowing %v on %s/%s: %v", a.GetOperation(), a.GetNamespace(), a.GetName(), newRoute)
+	return nil
+}
+
+var setExternalKubeInformerFactoryOnce sync.Once
+
+func (o *requiredRouteAnnotations) SetExternalKubeInformerFactory(kubeInformers informers.SharedInformerFactory) {
+	klog.Info("SetExternalKubeInformerFactory was called")
+	setExternalKubeInformerFactoryOnce.Do(func() {
```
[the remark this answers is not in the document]
candita: I noticed that we don't actually need to call `var _ = initializer.WantsExternalKubeInformerFactory(&requiredRouteAnnotations{})` because it is called automatically.  Removed it and removed the check here.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	if err = isRouteHSTSAllowed(ingress, newRoute, namespace); err != nil {
+		return admission.NewForbidden(a, err)
+	}
+	klog.Infof("allowing %v on %s/%s: %v", a.GetOperation(), a.GetNamespace(), a.GetName(), newRoute)
+	return nil
+}
+
+var setExternalKubeInformerFactoryOnce sync.Once
```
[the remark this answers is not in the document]
candita: no longer needed

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+
+		// Skip the validation if we're not making a change to HSTS  this  update
+		if wants == has && newhsts == oldhsts {
+			klog.Infof("ignoring %v on %s/%s that doesn't change HSTS policy (old=%v, new=%v)", a.GetOperation(), a.GetNamespace(), a.GetName(), oldhsts, newhsts)
+			return nil
+		}
+	}
+
+	// Since we are modifying route HSTS, wait for all caches to sync
+	if !o.waitForSyncedStore(ctx) {
```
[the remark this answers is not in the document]
candita: done, thanks!

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+		synced := []string{}
+		for i := range o.cachesToSync {
+			synced = append(synced, fmt.Sprintf("%v=%v", o.cachesToSyncNames[i], o.cachesToSync[i]()))
+		}
+		return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized: "+strings.Join(synced, ", ")))
+	}
```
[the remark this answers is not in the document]
candita: fixed, thanks!

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
sttts: objects in admission have internal types, not v1.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+	if a.GetOperation() == admission.Update {
+		wants, has := false, false
+		var oldhsts, newhsts string
+
+		newmetadata, err := meta.Accessor(newobject)
+		if err != nil {
+			return err
+		}
+		newannotations := newmetadata.GetAnnotations()
+		if newannotations != nil {
```
sttts, line 95: no need for nil check

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+			}
+			oldannotations := oldmetadata.GetAnnotations()
+			if oldannotations != nil {
+				oldhsts, has = oldannotations[hstsAnnotation]
+			}
+		}
+
+		// Skip the validation if we're not making a change to HSTS  this  update
+		if wants == has && newhsts == oldhsts {
+			klog.Infof("ignoring %v on %s/%s that doesn't change HSTS policy (old=%v, new=%v)", a.GetOperation(), a.GetNamespace(), a.GetName(), oldhsts, newhsts)
```
sttts, line 113: request level log output without loglevel is not good. at least make this V(4)

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+		return admission.NewForbidden(a, err)
+	}
+	klog.Infof("allowing %v on %s/%s: %v", a.GetOperation(), a.GetNamespace(), a.GetName(), newRoute)
+	return nil
+}
+
+var setExternalKubeInformerFactoryOnce sync.Once
+
+func (o *requiredRouteAnnotations) SetExternalKubeInformerFactory(kubeInformers informers.SharedInformerFactory) {
+	klog.Info("SetExternalKubeInformerFactory was called")
```
sttts, line 149: debug code

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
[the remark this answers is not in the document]
sttts: the reason for `var _ = type(x)` is not to call anything, but to assert an interface on `x`. If we change the interface, this assertion makes sure we notice that we suddenly don't implement the interface anymore. So please add the assertion again. It's our security net.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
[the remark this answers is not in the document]
sanchezl: This isn't satisfying any interface, just `return NewRequiredRouteAnnotations(), nil` in your actual `Factory` (the function literal inside your call to `plugins.Registry`).

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
sttts: Actually, I was talking about the group-kind. This is group-resource. Might be that that value is versioned. Wondering whether you really want the resource and not the kind though.

sanchezl: No, just here, use the group level API instead (e.g. import `"github.com/openshift/api/route"`).

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: Will do everywhere.  Thanks.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,369 @@
...
+	newRoute := a.GetObject().(*routeapi.Route)
+	namespace, err := o.nsLister.Get(newRoute.Namespace)
+	if err != nil {
+		return admission.NewForbidden(a, err)
+	}
+
+	if err = isRouteHSTSAllowed(ingress, newRoute, namespace); err != nil {
+		return admission.NewForbidden(a, err)
+	}
+	klog.Infof("allowing %v on %s/%s: %v", a.GetOperation(), a.GetNamespace(), a.GetName(), newRoute)
```
sttts, line 144: debug code

pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go
```diff
@@ -0,0 +1,512 @@
...
+		return true, fakeRoute(name, namespace, spec, map[string]string(routeAndAnnotations[name])), nil
+	}
+}
+
+func configFn(config *configv1.Ingress) clientgotesting.ReactionFunc {
+	return func(action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
+		return true, config, nil
+	}
+}
+*/
```
sttts, line 512: debug code

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,369 @@
...
+		klog.Infof("ignoring %v on %s/%s", a.GetOperation(), a.GetNamespace(), a.GetName())
+		return nil
+	}
+
+	// Determine if there are HSTS changes in this update
+	if a.GetOperation() == admission.Update {
+		wants, has := false, false
+		var oldhsts, newhsts string
+
+		newmetadata, err := meta.Accessor(newObject)
```
sanchezl, line 90: In general, use the expected convention of mixedCaps and initialisms for all the vars in this method. e.g:
`oldhsts` -> `oldHSTS`
`newmetadata` -> `newMetadata`
etc...

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: Sorry. All removed in next push.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,369 @@
...
+	o.syncCachesOnce.Do(func() {
+		o.cachesSynced = o.waitForSyncedStore(ctx)
+	})
+	if !o.cachesSynced {
+		synced := []string{}
+		for _, n := range o.cachesToSyncNames {
+			synced = append(synced, fmt.Sprintf("%s", n))
+		}
+		return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized: "+strings.Join(synced, ", ")))
+	}
```
sanchezl, line 128: We should keep checking until it _succeeds_ once.
```suggestion
	if !o.syncCachesSynced {
		o.cachesSynced = o.waitForSyncedStore(ctx)
	}
	if !o.syncCachesSynced {
		return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized: "))
	}
	// all good, continue....
```
Nit: From what I've seen in the logs, we can tell which caches did not sync from the logs output by the informers.

sanchezl (changes requested):

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: https://github.com/openshift/openshift-apiserver/pull/224/files#diff-cde26a88c8aa9772d3d1cbf96497ac0009504bc258fd662e2ad19b458dcd3d81R37

We are waiting for 10 seconds for it to succeed, but we don't want to wait forever in `Validate`.  Is that acceptable?

candita:
e2e-aws-upgrade failures coming from
 alert KubeAPIErrorBudgetBurn fired for 210 seconds with labels: {long="6h", severity="critical", short="30m"}

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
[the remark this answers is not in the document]
candita: ok, thanks.

candita:
Cluster operator authentication Degraded is True with IngressStateEndpoints_MissingSubsets::OAuthServerConfigObservation_Error::OAuthServerServiceEndpointAccessibleController_SyncError::OAuthServerServiceEndpointsEndpointAccessibleController_SyncError::RouterCerts_NoRouterCertSecret: IngressStateEndpointsDegraded: No subsets found for the endpoints of oauth-server

/test e2e-aws-upgrade

frobware:
/retitle "Bug 1986148: NE-310 HSTS Route Admission Plugin"

frobware:
/retitle Bug 1986148: NE-310 HSTS Route Admission Plugin

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -76,48 +79,47 @@ func (o *requiredRouteAnnotations) Validate(ctx context.Context, a admission.Att
...
+	// Wait for all caches to sync at least once
+	o.syncCachesOnce.Do(func() {
+		o.cachesSynced = o.waitForSyncedStore(ctx)
+	})
+	if !o.cachesSynced {
 		synced := []string{}
-		for i := range o.cachesToSync {
-			synced = append(synced, fmt.Sprintf("%v=%v", o.cachesToSyncNames[i], o.cachesToSync[i]()))
+		for _, n := range o.cachesToSyncNames {
+			synced = append(synced, fmt.Sprintf("%s", n))
```
frobware, line 122: We don't need this loop; instead this can be:

```go
	if !o.cachesSynced {
		return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized: "+strings.Join(o.cachesToSyncNames, ", ")))
	}
```

pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go
```diff
@@ -0,0 +1,495 @@
...
+	ingress := &configv1.Ingress{}
+	ingress.ObjectMeta.Name = name
+	ingress.Spec = config.Spec
+	return ingress
+}
+
+func fakeNamespaceLister(namespacesAndLabels map[string]map[string]string) corev1listers.NamespaceLister {
+	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+	for namespace, labels := range namespacesAndLabels {
+		indexer.Add(fakeNamespace(namespace, labels))
```
frobware, line 476: It may be benign but we should just that `Add()` was successful. Consider adding: 

```diff
From d4658691a41d4cb72b3e7dd99dcc1d70875e2d9e Mon Sep 17 00:00:00 2001
From: Andrew McDermott <amcdermo@redhat.com>
Date: Tue, 27 Jul 2021 10:17:40 +0100
Subject: [PATCH] Check for Index.Add() failures

---
 .../requiredrouteannotations/admission_test.go       | 12 +++++++++---
 1 file changed, 9 insertions(+), 3 deletions(-)

diff --git a/pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go b/pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go
index 690012d1e..1922e954f 100644
--- a/pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go
+++ b/pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go
@@ -473,14 +473,18 @@ func fakeIngress(name string, config configv1.Ingress) *configv1.Ingress {
 func fakeNamespaceLister(namespacesAndLabels map[string]map[string]string) corev1listers.NamespaceLister {
 	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	for namespace, labels := range namespacesAndLabels {
-		indexer.Add(fakeNamespace(namespace, labels))
+		if err := indexer.Add(fakeNamespace(namespace, labels)); err != nil {
+			panic(err.Error())
+		}
 	}
 	return corev1listers.NewNamespaceLister(indexer)
 }
 
 func fakeIngressLister(name string, config configv1.Ingress) configv1listers.IngressLister {
 	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
-	indexer.Add(fakeIngress(name, config))
+	if err := indexer.Add(fakeIngress(name, config)); err != nil {
+		panic(err.Error())
+	}
 	return configv1listers.NewIngressLister(indexer)
 }
 
@@ -488,7 +492,9 @@ func fakeRouteLister(routeDetails map[string]map[string]string, spec *routeapi.R
 	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	for rt, namespaceAndLabels := range routeDetails {
 		for ns := range namespaceAndLabels {
-			indexer.Add(fakeRoute(rt, ns, spec, namespaceAndLabels))
+			if err := indexer.Add(fakeRoute(rt, ns, spec, namespaceAndLabels)); err != nil {
+				panic(err.Error())
+			}
 		}
 	}
 	return routev1listers.NewRouteLister(indexer)
-- 
2.31.1
```

frobware (changes requested):
Overall LGTM but we need a test case to cover `SmallestMaxAge`. I did not see this line covered when running `go test` using: 

```bash
$ type cover
cover is a function
cover () 
{ 
    local t="/tmp/go-cover.$$.tmp";
    go test -coverprofile=$t $@ && go tool cover -html=$t
}
```

```go
func (c *hstsConfig) meetsRequirements(requirement configv1.RequiredHSTSPolicy) error {
        ...
        if requirement.MaxAge.SmallestMaxAge != nil && c.maxAge < *requirement.MaxAge.SmallestMaxAge {
                return fmt.Errorf("is less than minimum age (%d)", *requirement.MaxAge.SmallestMaxAge)
        }
```



- We need to follow up with end to end testing.

- I also would like to understand what happens on upgrade where the cluster has, say, 10,000 routes. From what I can see `Validate()` will back off until the caches have sync'd, but what will continue to call `Validate()` for all the existing routes where we return early (because we are not sync'd)?

pkg/route/apiserver/admission/requiredrouteannotations/admission_test.go
candita: Done, thanks.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: I've just taken care of these, thanks for the feedback.

Validate will block for up to 10 seconds the first time it is called to get caches sync'd.  It only needs to do that once.  In my manual testing, it logs less than 1/10th of a second to do that the first time (with only default routes).  I will do some further testing to provide an answer on upgrade with 10,000 routes.  I was (perhaps naively) relying on the admission framework to take care of this as a separation of concerns.

frobware: > Validate will block for up to 10 seconds the first time it is called to get caches sync'd.

Where is this 10 second specified? As part of the way sync works internally/

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: Spoke with @frobware and I understand now.  Will post a change.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: As discussed, will change to 30 seconds.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,365 @@
...
+		return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized: "+strings.Join(synced, ", ")))
+	}
+
+	ingress, err := o.ingressLister.Get("cluster")
+	if err != nil {
+		return admission.NewForbidden(a, err)
+	}
+
+	newRoute := a.GetObject().(*routeapi.Route)
+	namespace, err := o.nsLister.Get(newRoute.Namespace)
```
frobware, line 127: Isn't this redundant? Will `namespace` be different to `newRouter.Namespace`?

frobware: Ahh, are we just confirming the namespace actually exists?

candita: And nsLister has rights and is setup to access to it.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
frobware: > As discussed, will change to 30 seconds.

I really don't know whether this is too much. OTOH, we have customers with 10,000 routes and I don't know how long that will take to sync on a loaded cluster. Given that it will be a one-time affair, I'd rather be a little more conservative vis-a-vis it always fails because 10 seconds was never going to be enough. Would be interested in some feedback here from @sttts @sanchezl.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+		var oldHSTS, newHSTS string
+
+		newMetaData, err := meta.Accessor(newObject)
+		if err != nil {
+			return err
+		}
+		newAnnotations := newMetaData.GetAnnotations()
+		newHSTS, wants = newAnnotations[hstsAnnotation]
+
+		oldObject := a.GetOldObject()
```
sanchezl, line 95: Nit: consider using type assertion here to simplify the code below.
```suggestion
		oldObject := a.GetOldObject().(*routeapi.Route)
```

sanchezl (changes requested):

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+	if !isRoute {
+		return nil
+	}
+
+	// Skip the validation if we're not updating/creating
+	switch a.GetOperation() {
+	case admission.Update, admission.Create:
+	default:
+		return nil
+	}
```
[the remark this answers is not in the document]
candita: removed, thanks

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+	if a.GetOperation() == admission.Update {
+		wants, has := false, false
+		var oldHSTS, newHSTS string
+
+		newMetaData, err := meta.Accessor(newObject)
+		if err != nil {
+			return err
+		}
+		newAnnotations := newMetaData.GetAnnotations()
+		newHSTS, wants = newAnnotations[hstsAnnotation]
```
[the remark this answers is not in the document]
candita: removed, thanks

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+
+		oldObject := a.GetOldObject()
+		if oldObject != nil {
```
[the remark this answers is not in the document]
candita: fixed, thanks

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: done, thanks

candita:
@sanchezl I rebased the commits per your request.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,331 @@
...
+			return nil
+		}
+	}
+
+	// Wait up to 30 seconds for all caches to sync.  This is needed only once.
+	if !o.cachesSynced {
+		if synced := o.waitForSyncedStore(ctx); !synced {
+			return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized"))
+		}
+		o.cachesSynced = true
```
sttts, line 97: this races

sttts: needs locking or an atomic datatype

candita: Discussed with @sanchezl - removing the boolean, and now checking every time.

sanchezl (approved):
/lgtm
/retest

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: After a discussion with @sanchezl , and new feedback from @sttts , we decided to do this every time, and keep it 30 seconds.  30 seconds is a worst case scenario.  If needed, I can call a meeting with all of us together to hash it out.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
frobware: We could do the following:

```diff
diff --git a/pkg/route/apiserver/admission/requiredrouteannotations/admission.go b/pkg/route/apiserver/admission/requiredrouteannotations/admission.go
index f6b2f9279..268ab3fb0 100644
--- a/pkg/route/apiserver/admission/requiredrouteannotations/admission.go
+++ b/pkg/route/apiserver/admission/requiredrouteannotations/admission.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -50,7 +51,7 @@ type requiredRouteAnnotations struct {
 	routeLister   routev1listers.RouteLister
 	nsLister      corev1listers.NamespaceLister
 	ingressLister configv1listers.IngressLister
-	cachesSynced  bool
+	cachesSynced  cacheSync
 	cachesToSync  []cache.InformerSynced
 }
 
@@ -62,6 +63,24 @@ var _ = openshiftapiserveradmission.WantsOpenShiftRouteInformers(&requiredRouteA
 
 var maxAgeRegExp = regexp.MustCompile(`max-age=(\d+)`)
 
+type cacheSync struct {
+	readMu    sync.RWMutex
+	writeMu   sync.RWMutex
+	hasSynced bool
+}
+
+func (cs *cacheSync) HasSycned() bool {
+	cs.readMu.Lock()
+	defer cs.readMu.Unlock()
+	return cs.hasSynced
+}
+
+func (cs *cacheSync) SetSynced() {
+	cs.writeMu.Lock()
+	cs.hasSynced = true
+	cs.writeMu.Unlock()
+}
+
 // Validate ensures that routes specify required annotations, and returns nil if valid.
 // The admission handler ensures this is only called for Create/Update operations.
 func (o *requiredRouteAnnotations) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) (err error) {
@@ -90,11 +109,11 @@ func (o *requiredRouteAnnotations) Validate(ctx context.Context, a admission.Att
 	}
 
 	// Wait up to 30 seconds for all caches to sync.  This is needed only once.
-	if !o.cachesSynced {
-		if synced := o.waitForSyncedStore(ctx); !synced {
+	if !o.cachesSynced.HasSycned() {
+		if !o.waitForSyncedStore(ctx) {
 			return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized"))
 		}
-		o.cachesSynced = true
+		o.cachesSynced.SetSynced()
 	}
 
 	ingress, err := o.ingressLister.Get("cluster")
```

or more discretely with locking on the read and write access. I tried to encapsulate here to avoid any mistakes.

frobware:
@candita, @sttts mentioned to me that the we only have 30s for admission calls to complete so let's go for 20s on the sync. Sorry for the churn.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
frobware: fixing my typo;

```
func (cs *cacheSync) HasSynced() bool {
```

candita: I made the funcs private and changed a few other minor things.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+	cs.isSynced = true
+}
+
+type requiredRouteAnnotations struct {
+	*admission.Handler
+	routeLister   routev1listers.RouteLister
+	nsLister      corev1listers.NamespaceLister
+	ingressLister configv1listers.IngressLister
+	cachesToSync  []cache.InformerSynced
+	cacheSyncLock cacheSync
```
sttts, line 74:
```suggestion
        <empty line>
        isSyncedLock sync.RWMutex
	isSynced bool
```

sttts: And delete the data type above

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+	}
+
+	// Wait just once up to 20 seconds for all caches to sync
+	if !o.cacheSyncLock.hasSynced() {
+		if !o.waitForSyncedStore(ctx) {
+			return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized"))
+		}
+		o.cacheSyncLock.setSynced()
+	}
```
sttts, line 118:
```suggestion
        isSyncedLock.RLock()
        isSynced = o.isSynced
        isSyncedLock.RUnlock()
        if !isSynched {
		if !o.waitForSyncedStore(ctx) {
			return admission.NewForbidden(a, errors.New(pluginName+": caches not synchronized"))
		}
		isSyncedLock.Lock()
		o.isSynced = true
		isSyncedLock.Unlock()
	}
```

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,352 @@
...
+	o.cachesToSync = append(o.cachesToSync, kubeInformers.Core().V1().Namespaces().Informer().HasSynced)
+}
+
+// waitForSyncedStore calls cache.WaitForCacheSync, which will wait up to timeToWaitForCacheSync
+// for the cachesToSync to synchronize.
+func (o *requiredRouteAnnotations) waitForSyncedStore(ctx context.Context) bool {
+	syncCtx, cancelFn := context.WithTimeout(ctx, timeToWaitForCacheSync)
+	defer cancelFn()
+	return cache.WaitForCacheSync(syncCtx.Done(), o.cachesToSync...)
+}
```
sttts, line 147:
```suggestion
func (o *requiredRouteAnnotations) waitForSyncedStore(ctx context.Context) bool {
	        isSyncedLock.RLock()
	        defer isSyncedLock.RLock()
	        if o.isSynced {
	           return nil
	        }
	        
	syncCtx, cancelFn := context.WithTimeout(ctx, timeToWaitForCacheSync)
	defer cancelFn()
	err := cache.WaitForCacheSync(syncCtx.Done(), o.cachesToSync...)
	if err != nil {
	  return err
	 }
	 
	 o.isSyncedLock.Lock()
	 defer o.isSyncedLock.Unlock()
	 o.isSynced = true
	 
	 return err
}
```

candita: I made a similar change, ptal and let me know if you approve by approving. :)

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
```diff
@@ -0,0 +1,355 @@
...
+
+func (o *requiredRouteAnnotations) SetExternalKubeInformerFactory(kubeInformers informers.SharedInformerFactory) {
+	o.nsLister = kubeInformers.Core().V1().Namespaces().Lister()
+	o.cachesToSync = append(o.cachesToSync, kubeInformers.Core().V1().Namespaces().Informer().HasSynced)
+}
+
+// waitForSyncedStore calls cache.WaitForCacheSync, which will wait up to timeToWaitForCacheSync
+// for the cachesToSync to synchronize.
+func (o *requiredRouteAnnotations) waitForSyncedStore(ctx context.Context) bool {
+	syncCtx, cancelFn := context.WithTimeout(ctx, timeToWaitForCacheSync)
```
frobware, line 141: Don't know whether we're just endlessly bike shedding now, particularly if this is temporary and we're gravitating to a more general solution, but creating the context should move insider the if below. Is there any value in creating this each and every call?

candita: I will take this as a followup, per our discussion.

pkg/route/apiserver/admission/requiredrouteannotations/admission.go
candita: per our chat, this is a moot point

PR #225, opened 2021-06-29
Cleaning up some leftover packages

coreydaley:
Not sure how these got left in the go.sum, might need to add some additional validations


coreydaley:
/assign @sttts 
Looks like `go.sum` wasn't cleaned up when we got my last pull request merged

PR #227, opened 2021-07-07
Bug 1970805: Allow forward slashes in image name

coreydaley:
BuildConfigs should allow forward slashes `/` in the docker image name

https://bugzilla.redhat.com/show_bug.cgi?id=1970805


bparees:
/hold

you can't just allow slashes in imagestream names, that's got far reaching implications on the internal registry as well as on how we disambiguate the namespace and the imagestream name.

The easier fix here is probably to just change oc new-app/new-build so that when it generates a new imagestream, it replaces slashes with something else, to make the name valid.

coreydaley:
/close
This pull request is being closed in favor of a future pull request following the guidance in https://github.com/openshift/openshift-apiserver/pull/227#issuecomment-875801772

PR #229, opened 2021-07-08
Remove Dockerhub image test

lmzuccarelli:
- As we have a quay.io test, this is redundant
- Its causing errors in the CI pipeline due to docker's new rate limiting process


_output/named-golang-versions
```diff
@@ -0,0 +1,3 @@
+images/Dockerfile.rhel: 1.16
```
sgreene570, line 1: Is it necessary to commit the artifacts in `_output`?

lmzuccarelli: I will remove the folder. Thanks for pointing it out

sttts:
This need review from the owning team first.

sttts:
Please find somebody in the test owning team first for lgtm. Maybe @dmage ?

PR #230, opened 2021-07-14
Bug 1978627: Validate secret name and configMap name

coreydaley:
Validate the secret name and configMap name to provide a better user
experience when creating a build config that contains buildVolumes

Also, updating .gitignore to ignore _output directory that is not
cleaned up after running make verify

### DIFF
--- a/pkg/build/apis/build/validation/validation.go
+++ b/pkg/build/apis/build/validation/validation.go
@@ -810,9 +810,23 @@ func validateBuildVolume(volume buildapi.BuildVolume, fldPath *field.Path) field
 	var foundVolumeSources []string
 
 	if volume.Source.Secret != nil {
+		if len(volume.Source.Secret.SecretName) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("source").Child("secret").Child("secretName"), ""))
+		} else if reasons := validation.ValidateSecretName(volume.Source.Secret.SecretName, false); len(reasons) != 0 {
+			for _, r := range reasons {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("secret").Child("secretName"), volume.Source.Secret, r))
+			}
+		}
 		foundVolumeSources = append(foundVolumeSources, string(buildapi.BuildVolumeSourceTypeSecret))
 	}
 	if volume.Source.ConfigMap != nil {
+		if len(volume.Source.ConfigMap.Name) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("source").Child("configMap").Child("name"), ""))
+		} else if reasons := validation.ValidateConfigMapName(volume.Source.ConfigMap.Name, false); len(reasons) != 0 {
+			for _, r := range reasons {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("configMap").Child("name"), volume.Source.ConfigMap, r))
+			}
+		}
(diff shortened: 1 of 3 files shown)


adambkaplan (approved):
/lgtm

pkg/build/apis/build/validation/validation.go
```diff
@@ -810,9 +810,21 @@ func validateBuildVolume(volume buildapi.BuildVolume, fldPath *field.Path) field
 	var foundVolumeSources []string
 
 	if volume.Source.Secret != nil {
+		if len(volume.Source.Secret.SecretName) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("source").Child("secret").Child("secretName"), ""))
+		}
+		if reasons := validation.ValidateSecretName(volume.Source.Secret.SecretName, false); len(reasons) != 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("secret").Child("secretName"), volume.Source.Secret, "must be valid secret name"))
+		}
```
sttts, line 818: is this API new in 4.9 ? Or are there objects out in the wild on customer clusters?

sttts:
/hold

until https://github.com/openshift/openshift-apiserver/pull/230/files#r670400601 is answer. We cannot make validation stricter (without ratcheting validation)

pkg/build/apis/build/validation/validation.go
coreydaley: This is new

sttts:
> Also, updating .gitignore to ignore _output directory that is not
cleaned up after running make verify

Please don't mix these into other commits, even that trivial things. Make another commit.

pkg/build/apis/build/validation/validation.go
```diff
@@ -810,9 +810,21 @@ func validateBuildVolume(volume buildapi.BuildVolume, fldPath *field.Path) field
 	var foundVolumeSources []string
 
 	if volume.Source.Secret != nil {
+		if len(volume.Source.Secret.SecretName) == 0 {
+			allErrs = append(allErrs, field.Required(fldPath.Child("source").Child("secret").Child("secretName"), ""))
+		}
+		if reasons := validation.ValidateSecretName(volume.Source.Secret.SecretName, false); len(reasons) != 0 {
```
sttts, line 816: won't you get two errors now? When length is zero you don't have to call `ValidateSecretName`.

pkg/build/apis/build/validation/validation.go
```diff
@@ -810,9 +810,21 @@ func validateBuildVolume(volume buildapi.BuildVolume, fldPath *field.Path) field
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("source").Child("secret").Child("secretName"), volume.Source.Secret, "must be valid secret name"))
```
sttts, line 817: does it make sense to ignore `reasons`?

pkg/build/apis/build/validation/validation.go
sttts: To back this up: I think we run into `const dns1123LabelFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"` and an empty string does not match this.

coreydaley: updated in next push

pkg/build/apis/build/validation/validation.go
coreydaley: updated in next push

coreydaley:
@sttts ptal, comments have been addressed

PR #238, opened 2021-07-22
Bug 1971332: revert incorrect ssh scp fix

gabemontero:
we were just wrong with https://github.com/openshift/openshift-apiserver/pull/147

/assign @adambkaplan 

note with the redhat outage tests will fail, even unit (as there is an attempt to access registry.access.redhat.com in one of the tests)

### DIFF
--- a/pkg/build/apis/build/validation/validation_test.go
+++ b/pkg/build/apis/build/validation/validation_test.go
@@ -258,7 +258,7 @@ func TestBuildValidationFailure(t *testing.T) {
 	}
 }
 
-func TestBuildValidationWithSCPStyledURL(t *testing.T) {
+func TestBuildValidationWithSSHSCPStyledURL(t *testing.T) {
 	build := &buildapi.Build{
 		ObjectMeta: metav1.ObjectMeta{Name: "", Namespace: ""},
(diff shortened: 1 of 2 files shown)


pkg/build/apis/build/validation/validation_test.go
```diff
@@ -258,7 +258,7 @@ func TestBuildValidationFailure(t *testing.T) {
 	}
 }
 
-func TestBuildValidationWithSCPStyledURL(t *testing.T) {
+func TestBuildValidationWithSSHSCPStyledURL(t *testing.T) {
```
adambkaplan, line 261: I don't see a test that verifies `git@github.com:sclorg/nodejs-ex` - can you please add one?

adambkaplan (changes requested):
Would like to see a test to verify we are OK with SCP style URLs.

pkg/build/apis/build/validation/validation_test.go
gabemontero: added / pushed @adambkaplan

adambkaplan (approved):
/lgtm

PR #239, opened 2021-08-05
BUILD-188: LastTriggeredImageID marked as deprecated

alicerum:
LastTriggeredImageID field of BuildConfig Spec is marked as deprecated
and should not be set or used anymore.


pkg/build/apis/build/types.go
```diff
@@ -1091,6 +1091,9 @@ type WebHookTrigger struct {
 type ImageChangeTrigger struct {
 	// LastTriggeredImageID is used internally by the ImageChangeController to save last
 	// used image ID for build
+	//
+	// Deprecated: LastTriggerImageID is deprecated, will not be populated anymore and
+	// should not be used
```
adambkaplan, line 1096: Godoc should be updated in openshift/api first, then vendored in.
We have code generators that update the godoc.

adambkaplan (changes requested):

gabemontero:
@alicerum @adambkaplan so that ICT test case I've mentioned in scrum a time or two ....in case this is not already in your queue based on the e2e-aws results in this PR (though I did not see a related openshift/origin PR with @alicerum as the author), I believe that the image registry ICT test case that https://github.com/openshift/origin/pull/26098 touches will have to be changed in some fashion similar to what I did in that PR, as it inspects the build.spec.lasttriggeredimageID field.

Of course feel free to unhold that PR, use the commit as a basis for a new PR, or totally redo that PR in a different fashion to get the same results.

alicerum:
@gabemontero oh I am currently working on e2e failing tests for BUILD-188, but there is still no PR. I will take a closer look at your PR

adambkaplan:
/retitle Bug 1986562: Stop updating LastTriggeredImageID in spec

PR #243, opened 2021-08-31
Bug 1977414: return correct reason and message from failed build

alicerum:
When `Build` fails during `buildconfiginstantiate` (binary build), `status.reason`
and `status.message` are taken from the older state of the `Build` object, and they
are empty.

Must be taken from the new object obtained from `watch`.


adambkaplan (approved):
/lgtm

Build suite failures are unrelated to this change.

PR #246, opened 2021-09-14
Bug 2004203: no longer abort ICT induced builds because of spec last triggered image ID

gabemontero:
https://github.com/openshift/openshift-controller-manager/pull/201 is the more critical of the changes but I thought these changes were still useful for posterity

/assign @adambkaplan 
/assign @alicerum

### DIFF
--- a/pkg/build/apiserver/buildgenerator/generator.go
+++ b/pkg/build/apiserver/buildgenerator/generator.go
@@ -370,13 +370,19 @@ func (g *BuildGenerator) updateImageTriggers(ctx context.Context, bc *buildv1.Bu
 	if from != nil {
 		requestTrigger, responseTrigger = findImageChangeTrigger(bc, from)
 	}
-	if triggeredBy != nil &&
-		//TODO we still update spec until deprecated field (drepcated in 4.8) is removed (presumably 4.9)
-		((requestTrigger != nil && requestTrigger.LastTriggeredImageID == triggeredBy.Name) ||
-			(responseTrigger != nil && responseTrigger.LastTriggeredImageID == triggeredBy.Name)) {
+	if triggeredBy != nil && (responseTrigger != nil && responseTrigger.LastTriggeredImageID == triggeredBy.Name) {
 		klog.V(2).Infof("Aborting imageid triggered build for BuildConfig %s/%s with imageid %s because the BuildConfig already matches this imageid", bc.Namespace, bc.Name, triggeredBy.Name)
 		return fmt.Errorf("build config %s/%s has already instantiated a build for imageid %s", bc.Namespace, bc.Name, triggeredBy.Name)
 	}
+	// NOTE:  with this current form, if a BC was created and triggered while the cluster was prior to 4.8, as we are
+	// ignoring the spec last triggered here, that in theory would result in a one time retrigger of the build with the same image, where we store
+	// the details in the status instead of the spec.  However, with OCM ICT for BC now checking both spec and status, such
+	// a request should never come.  OCM ICT for BC should only instantiate if both spec and status last triggered do
+	// not match the latest.  The alternative, if we assumed OCM ICT for BC did not do the right thing, is have some
+	// kludgy one time copy of the spec last triggered information to status, and update the BC, but without instantiating.
+	// Of course, this takes the assumption that OCM and apiserver in a healthy, supportable cluster, are at the same version.
+	// If one of them failed to upgrade, the cluster is degraded regardless.
+
 	// Update last triggered image id for all image change triggers
 	// reset the status field in case the list of ICTs change in the spec; we'll then repopulate all of them
 	bc.Status.ImageChangeTriggers = []buildv1.ImageChangeTriggerStatus{}
--- a/pkg/build/apiserver/buildgenerator/generator_test.go
+++ b/pkg/build/apiserver/buildgenerator/generator_test.go
@@ -298,14 +298,29 @@ func TestInstantiateWithImageTrigger(t *testing.T) {
 			},
 		}
 	}
+	pre48Trigger := func() []buildv1.BuildTriggerPolicy {
+		return []buildv1.BuildTriggerPolicy{
+			{
+				Type: buildv1.ImageChangeBuildTriggerType,
+				ImageChange: &buildv1.ImageChangeTrigger{
+					From: &corev1.ObjectReference{
+						Name: "image1:tag1",
+						Kind: "ImageStreamTag",
+					},
+					LastTriggeredImageID: "ref/image1:tag1",
+				},
+			},
+		}
+	}
 	tests := []struct {
 		name    string
 		reqFrom *corev1.ObjectReference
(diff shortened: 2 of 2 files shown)


gabemontero:
e2e-aws-builds, e2e-aws, e2e-cmd were non sig-builds fails

/retest

pkg/build/apiserver/buildgenerator/generator.go
```diff
@@ -370,13 +370,19 @@ func (g *BuildGenerator) updateImageTriggers(ctx context.Context, bc *buildv1.Bu
 	if from != nil {
 		requestTrigger, responseTrigger = findImageChangeTrigger(bc, from)
 	}
-	if triggeredBy != nil &&
-		//TODO we still update spec until deprecated field (drepcated in 4.8) is removed (presumably 4.9)
-		((requestTrigger != nil && requestTrigger.LastTriggeredImageID == triggeredBy.Name) ||
-			(responseTrigger != nil && responseTrigger.LastTriggeredImageID == triggeredBy.Name)) {
+	if triggeredBy != nil && (responseTrigger != nil && responseTrigger.LastTriggeredImageID == triggeredBy.Name) {
```
adambkaplan, line 373: Why are we dropping the check on the "requestTrigger"?

pkg/build/apiserver/buildgenerator/generator_test.go
```diff
@@ -306,6 +320,7 @@ func TestInstantiateWithImageTrigger(t *testing.T) {
 		statusTriggerIndex int // index of trigger in status that will be updated with the imagev1 id, if -1, no update expected
 		triggers           []buildv1.BuildTriggerPolicy
 		errorExpected      bool
+		oldBC              bool
```
adambkaplan, line 323: This is not sufficiently descriptive. I would use a name that indicates we are updating the spec:

```suggestion
		lastTriggeredIDInSpec   bool
```

pkg/build/apiserver/buildgenerator/generator_test.go
```diff
@@ -338,6 +353,17 @@ func TestInstantiateWithImageTrigger(t *testing.T) {
...
+		{
+			name: "pre 4.7 trigger",
+			reqFrom: &corev1.ObjectReference{
+				Kind: "ImageStreamTag",
+				Name: "image1:tag1",
+			},
+			specTriggerIndex:   0,
+			statusTriggerIndex: 0,
+			triggers:           pre48Trigger(),
+			oldBC:              true,
```
adambkaplan, line 365:
```suggestion
			lastTriggeredIDInSpec:              true,
```

pkg/build/apiserver/buildgenerator/generator_test.go
```diff
@@ -440,28 +466,29 @@ func TestInstantiateWithImageTrigger(t *testing.T) {
...
-					t.Errorf("%s: expected trigger at index %d to contain imageID %s", tc.name, i, imageID)
-				}
-				continue
-			}
-			// Ensure that other triggers are updated with the latest container imagev1 ref
-			if bc.Spec.Triggers[i].Type == buildv1.ImageChangeBuildTriggerType {
-				from := bc.Spec.Triggers[i].ImageChange.From
-				if from == nil {
-					from = buildutil.GetInputReference(bc.Spec.Strategy)
+		if !tc.oldBC {
```
adambkaplan, line 469:
```suggestion
		// In 4.9 LastTriggeredImageID is no longer populated in spec. However BuildConfigs from clusters prior to
		// 4.9 may have this field populated.
		if !tc.lastTriggeredIDInSpec {
```

pkg/build/apiserver/buildgenerator/generator_test.go
```diff
@@ -440,28 +466,29 @@ func TestInstantiateWithImageTrigger(t *testing.T) {
...
+			for i := range bc.Spec.Triggers {
+				if i == tc.specTriggerIndex {
+					// Verify that the trigger got updated
```
adambkaplan, line 472:
```suggestion
					// Verify that the trigger in spec is empty
```

pkg/build/apiserver/buildgenerator/generator_test.go
```diff
@@ -306,6 +320,7 @@ func TestInstantiateWithImageTrigger(t *testing.T) {
 		statusTriggerIndex int // index of trigger in status that will be updated with the imagev1 id, if -1, no update expected
```
adambkaplan, line 320: There is a TODO comment above here - please update it.

adambkaplan (changes requested):

pkg/build/apiserver/buildgenerator/generator.go
gabemontero: I tried to convey that in the comment below

What got us in the mess in CI included the requestTrigger check flagging things here .... long term, IMO we only want status checks resulting in returning errors on the api server side

But as I tried to convey in the comment below, and the PR description, the meets min for fixing this is *ONLY* https://github.com/openshift/openshift-controller-manager/pull/201 and this change is *NOT* required

But again, I think we should be minimizing the use of the spec field anywhere we can.

We just cannot avoid it in OCM.

Hopefully that clarifies my intent / viewpoint.

WDYT wrt how we proceed here?

gabemontero:
updates for specific changes pushed @adambkaplan .... we just need to finalize on your "why remove" question and my explanation / answer

pkg/build/apiserver/buildgenerator/generator.go
gabemontero: Thinking through this is certainly warranted as part of review @adambkaplan 

My logic is that:
- once https://github.com/openshift/openshift-controller-manager/pull/201 merges, and with it still inspecting the spec, it will know not to trigger
- wrt ICTs, the ICT controllers in the OCM are the only "clients" that will be posting to this API server EP

Hence, I believe the answer to your question is "no"

Now, if there are flaws in that logic I am missing, then absolutely, we punt on this PR, and continue forever to check both, like we are on the OCM side.

And to reiterate, without https://github.com/openshift/openshift-controller-manager/pull/201, this code as it currently is lead to the repeated triggering churn..  If we had this change in place beforehand, the churn would not have occurred.

But we still need the OCM change to handle the case where a trigger is in fact needed.

adambkaplan:
/hold

Based on the conversation in https://github.com/openshift/openshift-apiserver/pull/246#discussion_r713270699, this shouldn't merge until https://github.com/openshift/openshift-controller-manager/pull/201 lands.

pkg/build/apiserver/buildgenerator/generator.go
adambkaplan: This makes more sense in light of the OCM pull request. I've place a hold so that the controller manager changes land first.

adambkaplan (approved):
/lgtm

pkg/build/apiserver/buildgenerator/generator.go
gabemontero: cool / sounds good

PR #253, opened 2021-11-01
Remove deprecated AdmissionTimeout usage

ardaguclu:
This PR removes deprecated AdmissionTimeout(ref: https://github.com/openshift/library-go/blob/28651a37805330bf95b0f5f30ad7c8a89619aaca/pkg/apiserver/admission/admissiontimeout/decorator.go#L11) and uses context to pass
timeout values to Mutation and Validation interfaces.


pkg/cmd/openshift-apiserver/openshiftapiserver/config.go
```diff
@@ -176,7 +175,6 @@ func NewOpenshiftAPIConfig(config *openshiftcontrolplanev1.OpenShiftAPIServerCon
 	admissionOptions := genericapiserveroptions.NewAdmissionOptions()
 	admissionOptions.Decorators = admission.Decorators{
 		admission.DecoratorFunc(admissionmetrics.WithControllerMetrics),
-		admission.DecoratorFunc(admissiontimeout.AdmissionTimeout{Timeout: 13 * time.Second}.WithTimeout),
```
tkashem, line 179: https://github.com/openshift/library-go/blob/28651a37805330bf95b0f5f30ad7c8a89619aaca/pkg/apiserver/admission/admissiontimeout/decorator.go#L16-L22


this timeout covers all admission controllers and webhooks and all resources/verbs, and this PR changes the scope to protecting only `instantiate`. is this desirable?

tkashem:
@sttts can you clarify the DEPRECATED statement? I am assuming that the admission timeout should be derived from the request context - maybe a quarter of the overall request deadline, or did you want to actually remove the `AdmissionTimeout` filter from the chain?

https://github.com/openshift/library-go/blob/28651a37805330bf95b0f5f30ad7c8a89619aaca/pkg/apiserver/admission/admissiontimeout/decorator.go#L11-L14

```
// DEPRECATED: use the context of the admission handler instead.
type AdmissionTimeout struct {
	Timeout time.Duration
}
```

pkg/cmd/openshift-apiserver/openshiftapiserver/config.go
ardaguclu: Thanks for review @tkashem. You are right, current status of this PR covers only instantiate, I'm still trying to change other parts as well. It is better to put WIP label first.

ardaguclu:
> @sttts can you clarify the DEPRECATED statement? I am assuming that the admission timeout should be derived from the request context - maybe a quarter of the overall request deadline, or did you want to actually remove the `AdmissionTimeout` filter from the chain?
> 
> https://github.com/openshift/library-go/blob/28651a37805330bf95b0f5f30ad7c8a89619aaca/pkg/apiserver/admission/admissiontimeout/decorator.go#L11-L14
> 
> ```
> // DEPRECATED: use the context of the admission handler instead.
> type AdmissionTimeout struct {
> 	Timeout time.Duration
> }
> ```

Good question, if it is not deprecated, I'll close PR. 

AFAIUI, if we start using context instead of admissiontimeout plugin, we have to put lots of places `context.withtimeout`

tkashem:
> AFAIUI, if we start using context instead of admissiontimeout plugin, we have to put lots of places `context.withtimeout`

i would like an approach where we don't have to make the change for every type. How about replacing `p.timeout` with a `context` that is derived from the parent `ctx` - we can say the admission timeout is 1/4th of the overall request deadline. If the ctx does not have any deadline, we can use a context with a timeout of `13s`.

This would be a change in library-go, but let's wait for @sttts to clarify :) 

https://github.com/openshift/library-go/blob/28651a37805330bf95b0f5f30ad7c8a89619aaca/pkg/apiserver/admission/admissiontimeout/timeoutadmission.go#L43

tkashem:
@ardaguclu I talked to Stefan. I think replacing `p.timeout` with a context timeout of `13s` is a good first step. let's forget about the "we can say the admission timeout is 1/4th of the overall request deadline" part for now.

can you please open a PR in library-go first with the context timeout fix? and then I think we have to re-vendor library-go in openshift/kubernetes and openshift/openshift-apiserver.

cc @sttts

ardaguclu:
> @ardaguclu I talked to Stefan. I think replacing `p.timeout` with a context timeout of `13s` is a good first step. let's forget about the "we can say the admission timeout is 1/4th of the overall request deadline" part for now.
> 
> can you please open a PR in library-go first with the context timeout fix? and then I think we have to re-vendor library-go in openshift/kubernetes and openshift/openshift-apiserver.
> 
> cc @sttts

Thanks @tkashem. I gladly open a PR in library-go.

ardaguclu:
This PR needs this https://github.com/openshift/library-go/pull/1239 to be re-vendored.

That's why, I'm closing and if the main one is merged, I can re-open it.

PR #248, opened 2021-09-24
Bug 1964112: Fix for host name validation error

lmzuccarelli:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1964112

A matrix regarding the alignment with the router can be found [here](https://docs.google.com/spreadsheets/d/16bY_SCk9hYQJE3G54y8WBk_nG2RWd5hMc3n1efc50YQ/edit#gid=0) 

The fix ensures alignment with the router if the annotation AllowNonDNSCompliantHostAnnotation is set to "false".
It enables the exact same validation mechanism as set in the openshift/router [here](https://github.com/openshift/router/blob/master/pkg/router/controller/unique_host.go#L240) 
To override this functionality set the annotation AllowNonDNSCompliantHostAnnotation to "true".

Validation steps :
 
Created 4.8.0 cluster (cluster-bot) with current PR
Deployed simple echo pod and service

Create Valid Route
- No annotations
- Name "poc-demo-test"
- curl -k -d'{"message":"hello"}' https://poc-demo-test-poc-demo.apps.ci-ln-zkkv962-f76d1.origin-ci-int-ce.dev.openshift.com/api/v1/echo - response {"Code":"200", "Sta    tus": "OK", "Message":"hello"}

Create Invalid Route
- No annotations
- Name (label greater than 63 chars) - "name-namespace-1234567890-1234567890-1234567890-12345678"
- Expected response - FAIL
- Router Status - NA
- The Route "name-namespace-1234567890-1234567890-1234567890-12345678" is invalid: spec.host: Invalid value: "name-namespace-1234567890-1234567890-1234567890-12345678-p    oc-demo.apps.ci-ln-zkkv962-f76d1.origin-ci-int-gce.dev.openshift.com": host (label) must conform to DNS 1123 label conventions must be no more than 63 characters

Create Valid Route (using annotation)
- Created route (label greater than 63 chars)
- Added annotation route.openshift.io/allow-non-dns-compliant-host: "true"
- Expected response - PASS (route created) from apiserver
- Router Status - Rejected
host name validation errors: spec.host: Invalid value: "name-namespace-1234567890-1234567890-1234567890-1234567-poc-demo.apps.ci-ln-zkkv962-f76d1.origin-ci-int-gce.de    v.openshift.com": must be no more than 63 characters

PR #210, opened 2021-05-24
Bug 1957584: Revert 'Bug 1896977: Enhance API host name validation'

Miciah:
This reverts commit 402c0fdc3b10abfef4d3b4fbc09e14e33a0a8dbf.

PR #254, opened 2021-11-02
NE-882: Allow route subdomains to be omitted

Miciah:
When a subdomain field is set, and host is unset, allow host to be empty.

---

This is https://github.com/openshift/openshift-apiserver/pull/194, rebased for https://github.com/openshift/openshift-apiserver/pull/210 and https://github.com/openshift/openshift-apiserver/pull/248.

I had planned to add `ValidateSubdomain` with https://github.com/openshift/library-go/pull/1240 and use that in this PR, but I got mired in a go-mod morass, gave up, and added a to-do comment to use use library-go in the future.

### DIFF
--- a/pkg/route/apis/route/validation/validation_test.go
+++ b/pkg/route/apis/route/validation/validation_test.go
@@ -536,6 +536,93 @@ func TestValidateRoute(t *testing.T) {
 			expectedErrors: 1,
 		},
 		{
+			name: "Valid subdomain",
+			route: &routeapi.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "foo",
+				},
+				Spec: routeapi.RouteSpec{
+					Subdomain: "api.ci",
+					To:        createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 0,
+		},
+		{
+			name: "Valid subdomain (253 chars)",
+			route: &routeapi.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "foo",
+				},
+				Spec: routeapi.RouteSpec{
+					Subdomain: "name-namespace.a1234567890.b1234567890.c1234567890.d1234567890.e1234567890.f1234567890.g1234567890.h1234567890.i1234567890.j1234567890.k1234567890.l1234567890.m1234567890.n1234567890.o1234567890.p1234567890.q1234567890.r1234567890.s12345678.test.com",
+					To:        createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 0,
+		},
+		{
+			name: "Invalid subdomain (279 chars)",
+			route: &routeapi.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "foo",
+				},
+				Spec: routeapi.RouteSpec{
+					Subdomain: "name-namespace.a1234567890.b1234567890.c1234567890.d1234567890.e1234567890.f1234567890.g1234567890.h1234567890.i1234567890.j1234567890.k1234567890.l1234567890.m1234567890.n1234567890.o1234567890.p1234567890.q1234567890.r1234567890.s1234567890.t1234567890.u1234567890.test.com",
+					To:        createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 1,
+		},
+		{
+			name: "Invalid DNS 952 subdomain",
+			route: &routeapi.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "name",
+					Namespace: "foo",
+				},
+				Spec: routeapi.RouteSpec{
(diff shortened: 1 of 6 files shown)


pkg/route/apis/route/validation/validation_test.go
```diff
@@ -535,6 +535,34 @@ func TestValidateRoute(t *testing.T) {
...
+				},
+				Spec: routeapi.RouteSpec{
+					Subdomain: "api.ci",
+					To:        createRouteSpecTo("serviceName", "Service"),
+				},
+			},
+			expectedErrors: 0,
+		},
+		{
+			name: "Invalid DNS 952 subdomain",
```
gcs278, line 553: I see we check for invalid characters in DNS name, would it make sense to also add one that tests `kvalidation.DNS1123SubdomainMaxLength`?

gcs278 (commented):
Overall looks good to me

pkg/route/apis/route/validation/validation_test.go
Miciah: I've added some more tests for subdomain.  That did lead me wonder whether we should restrict the subdomain to some maximum length less then 253 chars; after all, the subdomain is going to be prepended to the router's domain, which will be more then 0 chars.  However, we can't predict how long the router's domain will be, so maybe better not to pick an arbitrary value that will reject overly long subdomains in some cases and not in others.  What do you think?

gcs278: Oh that's right. Hmm. I'd think it's best to just restrict subdomain to the entire domain length and call it a day. Is there another length check for the entire domain, when we add subdomain to it? Is there warning/errors that we should throw if it's too long?

Miciah:
Last push includes the results of doing a gofmt.

Miciah:
Last push added warnings for routes with `spec.host` and `spec.subdomain` both set.

Miciah:
Rebased.

Miciah:
Here is example `oc create` output for the new warning.
<details>
<summary>Example output</summary>

```
% cat test1.yaml         
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: test1
spec:
  host: test1-e2e-test-router-subdomain-7nfk7.apps.ci-ln-hzmj2lb-72292.origin-ci-int-gce.dev.rhcloud.com
  to:
    name: test
% oc create -f test1.yaml
route.route.openshift.io/test1 created
% cat test2.yaml 
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: test2
spec:
  host: foo.tld
  to:
    name: test
% oc create -f test2.yaml
route.route.openshift.io/test2 created
% cat test3.yaml
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: test3
spec:
  subdomain: foo
  to:
    name: test
% oc create -f test3.yaml
route.route.openshift.io/test3 created
% cat test4.yaml 
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: test4
spec:
  host: foo.tld
  subdomain: foo
  to:
    name: test
% oc create -f test4.yaml
Warning: spec.host is set; spec.subdomain will be ignored
route.route.openshift.io/test4 created
% oc get routes
NAME    HOST/PORT                                                                                          PATH   SERVICES   PORT    TERMINATION   WILDCARD
test1   test1-e2e-test-router-subdomain-7nfk7.apps.ci-ln-hzmj2lb-72292.origin-ci-int-gce.dev.rhcloud.com          test       <all>                 None
test2   foo.tld                                                                                                   test       <all>                 None
test3   foo.apps.ci-ln-hzmj2lb-72292.origin-ci-int-gce.dev.rhcloud.com                                            test       <all>                 None
test4   HostAlreadyClaimed                                                                                        test       <all>                 None
```
</details>

Miciah:
I changed the warning from "will be ignored" to "may be ignored" because the API shouldn't assume that the router implements `spec.subdomain`.

Miciah:
> I changed the warning from "will be ignored" to "may be ignored" because the API shouldn't assume that the router implements `spec.subdomain`.

On second thought...  If `spec.subdomain` isn't implemented, then it will _definitely_ be ignored.  :thinking:...   ¯\\\_(ツ)\_/¯.  The important thing is that there's a warning.

gcs278:
/lgtm
:100:

Miciah:
```
level=error msg=Error: no matching EC2 Subnet found
level=error
level=error msg=  with module.vpc.data.aws_subnet.private[0],
level=error msg=  on vpc/common.tf line 22, in data "aws_subnet" "private":
level=error msg=  22: data "aws_subnet" "private" {
level=error
level=error msg=failed to fetch Cluster: failed to generate asset "Cluster": failure applying terraform for "cluster" stage: failed to create cluster: failed to apply Terraform: exit status 1
level=error
level=error msg=Error: no matching EC2 Subnet found
level=error
level=error msg=  with module.vpc.data.aws_subnet.private[0],
level=error msg=  on vpc/common.tf line 22, in data "aws_subnet" "private":
level=error msg=  22: data "aws_subnet" "private" { 
```
/test e2e-aws

PR #260, opened 2021-11-23
WIP Bug-1949870: Reference to a broken tag using ImageStreamTag does not …

Nikita757:
…reflect in ImageStream status.

Added some additional functionality in case if ImageStreamTags reference to a broken image, so they would also appear in status of an image stream.

https://bugzilla.redhat.com/show_bug.cgi?id=1949870


pkg/image/apiserver/internalimageutil/helpers.go
```diff
@@ -394,7 +412,7 @@ func AddTagEventToImageStream(stream *imageapi.ImageStream, tag string, next ima
 	return true
 }
 
-// UpdateTrackingTags sets updatedImage as the most recent TagEvent for all tags
+// UpdateTrackingTags sets updatedTagList as the most recent TagEventList for all tags
```
dmage, line 415: It's not enough to replace names and types in comments. What does the updated comment mean? It's also worth updating the example, what'll happen if `status.tags[2.0].items` is empty?

pkg/image/apiserver/internalimageutil/helpers.go
```diff
@@ -463,9 +486,16 @@ func UpdateTrackingTags(stream *imageapi.ImageStream, updatedTag string, updated
...
+				updated++
+			}
+		}
 
-		if AddTagEventToImageStream(stream, specTag, updatedImage) {
-			klog.V(5).Infof("stream updated")
+		if len(updatedTagList.Conditions) != 0 {
+			AddTagConditionToImageStream(stream, specTag, updatedTagList.Conditions[0])
+			klog.V(5).Infoln("stream updated with tag conditions")
 			updated++
```
dmage, line 499: The comment says _Returns the number of tags changed_, but it seems this patch changes semantics of the return value (I would say it shouldn't be changed though).

pkg/image/apiserver/internalimageutil/helpers.go
```diff
@@ -463,9 +486,16 @@ func UpdateTrackingTags(stream *imageapi.ImageStream, updatedTag string, updated
...
+			if AddTagEventToImageStream(stream, specTag, updatedTagList.Items[0]) {
+				klog.V(5).Infof("stream updated with tag events")
+				updated++
+			}
+		}
 
-		if AddTagEventToImageStream(stream, specTag, updatedImage) {
-			klog.V(5).Infof("stream updated")
+		if len(updatedTagList.Conditions) != 0 {
+			AddTagConditionToImageStream(stream, specTag, updatedTagList.Conditions[0])
```
dmage, line 497: It might be misleading if we just copy conditions.

If this logic didn't exist, and we had only ability only to import `DockerImage` and were importing another tag through the registry, the client will observe next situations:

(Let's assume we have a tag `2.0`, and a tag `latest` that points to `2.0`, i.e. that has `from.name: "2.0"`)

1. The tag 2.0 has a pushed image or a successfully imported image, i.e. it has something in `status.tags[2.0].items` and doesn't have conditions. In this case the image `status.tags[2.0].items[0]` can be pulled from `image-registry.openshift-image-registry.svc:5000/namespace/imagestream:2.0`. So `status.tags[latest].items[0]` should point to the same image and shouldn't have failed conditions.
2.  The tag 2.0 has an imported image (`status.tags[2.0].items` is not empty), but its spec has been updated and the new image cannot be imported (i.e. it also has conditions). The latest successfully imported image still ca be pulled from `image-registry.openshift-image-registry.svc:5000/namespace/imagestream:2.0`, so `status.tags[latest].items[0]` should point to the latest successfully imported image. We can ignore conditions on the tag `2.0` and the tag `latest` shouldn't have failed conditions.
3. The tag 2.0 exists (or doesn't exist), but it hasn't been imported yet. Perhaps it has conditions OR the image importer controller is too busy and hasn't seen the image stream.  If you try to pull the image from `image-registry.openshift-image-registry.svc:5000/namespace/imagestream:2.0`, the registry will say you `404 Not Found`. In this case `status.tags[latest]` should have conditions that the source doesn't exist. This condition should exist in both cases when the tag 2.0 has conditions or when it doest exist at all.

dmage (changes requested):

pkg/image/apiserver/registry/imagestream/strategy.go
```diff
@@ -319,17 +319,23 @@ func (s Strategy) tagsChanged(ctx context.Context, old, stream *imageapi.ImageSt
...
+		if len(newImages.Conditions) == 0 {
+			conditionsPresent = false
+		} else {
+			conditionsPresent = true
+		}
+		if len(newImages.Items) == 0 && !conditionsPresent {
 			continue
 		}
-		if old != nil {
+		if old != nil && !conditionsPresent {
```
dmage, line 331: What's this code doing?

PR #261, opened 2021-11-24
Bug 2022188: Image policy should mutate DeploymentConfigs

dmage:
This is a backport of #259 to address the [bug 2022188](https://bugzilla.redhat.com/show_bug.cgi?id=2022188).

**Impact:** Resolution of image stream names in deployment configs does not work because the image policy plugin does not know about them.

**Cause:** The problem exists since the beginning, but it wasn't visible as the image policy plugin used to resolve image streams names in all pods. Once we've fixed the [bug 1925180](https://bugzilla.redhat.com/show_bug.cgi?id=1925180), the plugin doesn't update controlled objects anymore, i.e. it doesn't update pods that are created for the deployment config.

**Resolution:** Update the admission plugin so that is aware of deployment configs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.2-0.20190823105129-775207bd45b6
 	github.com/openshift/api v0.0.0-20210115203145-3b72f192aace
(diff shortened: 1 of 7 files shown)


sttts:
Somebody from the team has to lgtm first.

xiuwang:
@wangke19 Could you help add /label cherry-pick-approved ? Thanks

PR #267, opened 2022-01-06
projects: add rw mutex to auth cache

sttts:
The auth cache of the projects storage implementation did not lock its data. This is most likely the source why projects sometimes did not show up, a bug we have heard from customers for long long time.

All credits to @davidfestal for finding this through his kcp work.

### DIFF
--- a/pkg/project/auth/cache_test.go
+++ b/pkg/project/auth/cache_test.go
@@ -1,14 +1,17 @@
 package auth
 
 import (
+	"context"
 	"fmt"
 	"strconv"
+	"sync"
 	"testing"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/authentication/user"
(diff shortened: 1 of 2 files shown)


s-urbaniak:
can we we add some test leveraging the fault using `go test -race`?

smarterclayton:
Yay!

pkg/project/auth/cache_test.go
```diff
@@ -175,3 +177,47 @@ func TestSyncNamespace(t *testing.T) {
...
+			nsIndexer.Add(&corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("foo%d", i), ResourceVersion: fmt.Sprintf("%d", i)},
+			})
+			i++
+			authorizationCache.synchronize()
+		}, 0)
+	}()
+
+	for i := 0; i < 1000; i++ {
+		authorizationCache.List(alice, labels.Everything())
```
s-urbaniak, line 212: Shouldn't this be:
```
	var wg sync.WaitGroup
	for i := 0; i < 1000; i++ {
		wg.Add(1)
		go func() {
			authorizationCache.List(alice, labels.Everything())
		}()
	}
	wg.Wait()
```

only then we can provoke the reader lock and concurrent access.

s-urbaniak:
lgtm, but leaving final eyes to @stlaz

ibihim:
lgtm

PR #275, opened 2022-01-31
Bug 2047335: fix panic from printer

tkashem:
Related https://github.com/openshift/openshift-apiserver/pull/272


pkg/security/printers/internalversion/printer.go
```diff
@@ -103,10 +103,10 @@ func printSecurityContextConstraint(scc *securityapi.SecurityContextConstraints,
 		scc.Name,
 		scc.AllowPrivilegedContainer,
 		strings.Join(capabilities, ","),
-		scc.SELinuxContext.Type,
-		scc.RunAsUser.Type,
```
p0lyn0mial, line 107: could we add a ut so that in the future (when this method/type will be extended) the issue is detected immediately?

p0lyn0mial:
also, I am wondering if there is a more systematic way to detect other potential places?

pkg/security/printers/internalversion/printer.go
tkashem: i think we should add unit tests for these `TableRow` printers.

p0lyn0mial:
> also, I am wondering if there is a more systematic way to detect other potential places?

should we write an e2e test that requires tabular representation of objects?

PR #279, opened 2022-02-10
Bug 2049234: Fix importing images that have dots in their namespace

dmage:
This PR bumps library-go to get a new registryclient that can correctly
handle image references like

    registry.example.com/namespace.with.dot/foo

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/openshift/apiserver-library-go v0.0.0-20211209162547-8c11dbc46b6e
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
 	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
-	github.com/openshift/library-go v0.0.0-20211216141749-1816407208cf
+	github.com/openshift/library-go v0.0.0-20220210170159-18f172cff934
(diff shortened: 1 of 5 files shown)


pkg/openapi/zz_generated.openapi.go
```diff
@@ -61602,11 +61602,11 @@ func schema_k8sio_api_core_v1_LimitRangeItem(ref common.ReferenceCallback) commo
...
+							Description: "Type of resource that this limit applies to.\n\nPossible enum values:\n - `\"Container\"` Limit that applies to all containers in a namespace\n - `\"PersistentVolumeClaim\"` Limit that applies to all persistent volume claims in a namespace\n - `\"Pod\"` Limit that applies to all pods in a namespace",
 							Default:     "",
 							Type:        []string{"string"},
 							Format:      "",
-							Enum:        []interface{}{"Container", "PersistentVolumeClaim", "Pod", "openshift.io/Image", "openshift.io/ImageStream"}},
```
dmage, line 61609: @tkashem any ideas why images are removed from this list?

dmage: hm... it was a strange flake

tkashem: can you please open a BZ with this flake, we should fix it.

PR #285, opened 2022-02-17
 [WIP] (DO NOT MERGE) Bug 2041133: Testing vendor update of API repo for fixing oc explain for route conditions #285

gcs278:
PR is for a test cluster of a full re-vendored API repo with a `make update`. I'm new to working with these repos and I'm trying to understand the upstream/downstream implications.

When I vendored in the latest API repo, a lot more changed than what I only changed, so I'm just interested in validating my change along with the other changes.

DO NOT MERGE. This PR is rev-ing the API repo and it has my hacked in changes (since my changes aren't pushed into the API repo yet so there is no SHA to reference..chicken and the egg..).


melvinjoseph86:
Verified via pre-merge verification workflow, more references related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=2041133#c3

PR #292, opened 2022-07-06
Bug 2089402: Validate Image labels correctly

jkhelil:


pkg/build/apis/build/validation/validation_test.go
```diff
@@ -2564,7 +2564,7 @@ func TestValidateCommonSpecSuccess(t *testing.T) {
 					ImageLabels: []buildapi.ImageLabel{
 						{
 							Name:  "key",
-							Value: "value )(*&",
+							Value: "value-_.v",
```
coreydaley, line 2567: Why did this need to change?

pkg/build/apis/build/validation/validation_test.go
```diff
@@ -3944,3 +3944,83 @@ func TestValidateBuildVolumes(t *testing.T) {
...
+				{
+					Name:  "key1",
+					Value: "value1",
+				},
+				{
+					Name:  "prefix/key2",
+					Value: "value2",
+				},
+			},
+		},
```
coreydaley, line 4003: Missing the explicit err fields.

jkhelil: we shouldnt have errors here, it is passing a test

pkg/build/apis/build/validation/validation_test.go
coreydaley: My concern is that if we make this more restrictive, all of the sudden we will be generating errors for existing labels that include those characters (if any exist)

pkg/build/apis/build/validation/validation.go
```diff
@@ -735,10 +735,10 @@ func ValidateImageLabels(labels []buildapi.ImageLabel, fldPath *field.Path) (all
...
 		if len(lbl.Name) == 0 {
 			allErrs = append(allErrs, field.Required(idxPath.Child("name"), ""))
-			continue
-		}
-		for _, msg := range kvalidation.IsConfigMapKey(lbl.Name) {
-			allErrs = append(allErrs, field.Invalid(idxPath.Child("name"), lbl.Name, msg))
+		} else if errs := kvalidation.IsQualifiedName(lbl.Name); len(errs) != 0 {
+			allErrs = append(allErrs, field.Invalid(idxPath.Child("name"), lbl.Name, strings.Join(errs, "; ")))
+		} else if errs := kvalidation.IsValidLabelValue(lbl.Value); len(errs) != 0 {
+			allErrs = append(allErrs, field.Invalid(idxPath.Child("value"), lbl.Value, strings.Join(errs, "; ")))
```
coreydaley, line 741: We were not previously checking the value, only the name, so let's not add the checking of the value, that should fix the `making the image label value more restrictive` issue and you should be able to revert the other test change.  If we get a bug about image values, then we will address it, but we probably won't be able to backport that change.

pkg/build/apis/build/validation/validation_test.go
coreydaley: Revert this and see above comment about having added the `IsValidLabelValue`.

pkg/build/apis/build/validation/validation_test.go
jkhelil: done

pkg/build/apis/build/validation/validation.go
jkhelil: Removed

pkg/build/apis/build/validation/validation_test.go
jkhelil: fixed, I removed the check for label values

bparees:
/approve

@coreydaley please work with the api-server team to get yourself added to https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/OWNERS (and perhaps remove the old/stale entries)

coreydaley:
@bparees I have opened https://github.com/openshift/openshift-apiserver/pull/294

PR #304, opened 2022-07-20
add retry on imagestreamtag update

neisw:
Adding retry during update to address test failures

https://search.ci.openshift.org/?search=.*Operation+cannot+be+fulfilled+on+imagestreamtags.*&maxAge=336h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.12-e2e-aws-sdn-serial/1548438013934047232

: [sig-imageregistry][Feature:ImageTriggers][Serial] ImageStream API TestImageStreamTagLifecycleHook [Suite:openshift/conformance/serial]

 Message: "Operation cannot be fulfilled on imagestreamtags.image.openshift.io \"test\": the object has been modified; please apply your changes to the latest version and try again",
 
Investigation indicated that the test was not specifying a particular resourceVersion, another call (outside the audit trail) was modifying the resource between the time is was fetched and then update called resulting in the failure.  Given that a specific resourceVersion was not provided the belief is that a retry is warranted in this case.

### DIFF
--- a/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
+++ b/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
@@ -24,6 +24,10 @@ func NewImageStreamRegistryTester(registry imagestream.Registry, apiTesters map[
 	tester.Registry = registry
 	tester.registryApiTesters = apiTesters
 
+	if tester.registryApiTesters == nil {
+		tester.registryApiTesters = make(map[string]*ApiTester)
+	}
+
 	return tester
 }
 
@@ -103,27 +107,30 @@ func (r *imageStreamRegistryTester) WatchImageStreams(ctx context.Context, optio
 
 func (r *imageStreamRegistryTester) extractImageStreamResponse(apiName string) (bool, *imageapi.ImageStream, error) {
 
-	if apiTester, ok := r.registryApiTesters[apiName]; ok {
+	if r.registryApiTesters != nil {
 
-		// increment the call count when done
-		defer apiTester.callComplete()
+		if apiTester, ok := r.registryApiTesters[apiName]; ok {
 
-		if responses, okResponses := apiTester.callResponses[apiTester.callCount]; okResponses {
+			// increment the call count when done
+			defer apiTester.callComplete()
 
-			var err error = nil
+			if responses, okResponses := apiTester.callResponses[apiTester.callCount]; okResponses {
 
-			if responseErr, okResponse := responses.response["error"]; okResponse {
-				err = responseErr.(error)
+				var err error = nil
+
+				if responseErr, okResponse := responses.response["error"]; okResponse {
+					err = responseErr.(error)
+				}
+
+				if responseImageStream, okImageStream := responses.response["imageStream"]; okImageStream {
+					imageStream := responseImageStream.(imageapi.ImageStream)
+					return true, &imageStream, err
+				} else {
+					return true, nil, err
+				}
 			}
 
-			if responseImageStream, okImageStream := responses.response["imageStream"]; okImageStream {
-				imageStream := responseImageStream.(imageapi.ImageStream)
-				return true, &imageStream, err
-			} else {
-				return true, nil, err
-			}
 		}
-
 	}
 
 	return false, nil, nil
--- a/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
+++ b/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
@@ -2,6 +2,7 @@ package imagestreamtag
 
 import (
 	"context"
+	"sync/atomic"
 
 	imageapi "github.com/openshift/openshift-apiserver/pkg/image/apis/image"
 	"github.com/openshift/openshift-apiserver/pkg/image/apiserver/registry/imagestream"
@@ -43,20 +44,20 @@ func NewApiResponse() ApiResponse {
 }
 
 type ApiTester struct {
-	callResponses map[int]ApiResponse
-	callCount     int
+	callResponses map[int32]ApiResponse
+	callCount     int32
 }
 
 func NewApiTester() *ApiTester {
 	tester := &ApiTester{}
 	tester.callCount = 0
-	tester.callResponses = make(map[int]ApiResponse)
+	tester.callResponses = make(map[int32]ApiResponse)
 
 	return tester
 }
 
 func (a *ApiTester) callComplete() {
-	a.callCount++
+	atomic.AddInt32(&a.callCount, 1)
 }
 
 func (r *imageStreamRegistryTester) ListImageStreams(ctx context.Context, options *metainternal.ListOptions) (*imageapi.ImageStreamList, error) {
--- a/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
+++ b/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
@@ -2,12 +2,15 @@ package imagestreamtag
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 
 	imageapi "github.com/openshift/openshift-apiserver/pkg/image/apis/image"
 	"github.com/openshift/openshift-apiserver/pkg/image/apiserver/registry/imagestream"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metainternal "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
 )
 
@@ -136,4 +139,14 @@ func (r *imageStreamRegistryTester) extractImageStreamResponse(apiName string) (
 
 	return false, nil, nil
 
+}
+
+func createInvalidError() error {
+	gk := schema.GroupKind{Group: "imageregistry.operator.openshift.io", Kind: "anyKind"}
+	return errors.NewInvalid(gk, "test", nil)
+}
+
+func createConflictError() error {
+	gr := schema.GroupResource{Group: "imageregistry.operator.openshift.io", Resource: "configs"}
+	return errors.NewConflict(gr, "test", fmt.Errorf("testing error"))
 }
--- a/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
+++ b/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
@@ -10,7 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 )
 
-type ImageStreamRegistryTester struct {
+type imageStreamRegistryTester struct {
 	// the true implementation
 	imagestream.Registry
 
@@ -18,8 +18,8 @@ type ImageStreamRegistryTester struct {
 	registryApiTesters map[string]*ApiTester
 }
 
-func NewImageStreamRegistryTester(registry imagestream.Registry, apiTesters map[string]*ApiTester) *ImageStreamRegistryTester {
-	tester := &ImageStreamRegistryTester{}
+func NewImageStreamRegistryTester(registry imagestream.Registry, apiTesters map[string]*ApiTester) *imageStreamRegistryTester {
+	tester := &imageStreamRegistryTester{}
 
 	tester.Registry = registry
 	tester.registryApiTesters = apiTesters
@@ -29,24 +29,24 @@ func NewImageStreamRegistryTester(registry imagestream.Registry, apiTesters map[
 
 // Takes a response variable name that will be cast to the proper response type
 type ApiResponse struct {
-	response map[string] interface{}
+	response map[string]interface{}
 }
 
 func NewApiResponse() ApiResponse {
 	response := ApiResponse{}
-	response.response = make(map[string] interface{})
+	response.response = make(map[string]interface{})
 	return response
 }
 
 type ApiTester struct {
-	callResponses map[int] ApiResponse
+	callResponses map[int]ApiResponse
 	callCount     int
 }
 
 func NewApiTester() *ApiTester {
 	tester := &ApiTester{}
 	tester.callCount = 0
-	tester.callResponses = make(map[int] ApiResponse)
+	tester.callResponses = make(map[int]ApiResponse)
 
 	return tester
 }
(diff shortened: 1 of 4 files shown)


pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -206,20 +207,44 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
+	})
+
 	if err != nil {
 		return nil, false, err
 	}
 
+	return result, created, updateErr
+}
+
+func (r *REST) update(ctx context.Context, tagName string, objInfo rest.UpdatedObjectInfo, createValidation rest.ValidateObjectFunc, updateValidation rest.ValidateObjectUpdateFunc, forceAllowCreate bool, options *metav1.UpdateOptions) (runtime.Object, bool, error, bool) {
```
deads2k, line 233: canonically error goes last

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -266,19 +294,20 @@ func (r *REST) Update(ctx context.Context, tagName string, objInfo rest.UpdatedO
 		istag.Labels = imageStream.Labels
 	}
 
+	// should we consider retry here or is this outside the scope of conflict checking?
 	if create {
 		if err := rest.BeforeCreate(r.strategy, ctx, obj); err != nil {
-			return nil, false, err
+			return nil, false, err, canRetry
```
deads2k, line 300: a create should never produce a 409, so this should be false in this branch.

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -308,18 +337,20 @@ func (r *REST) Update(ctx context.Context, tagName string, objInfo rest.UpdatedO
 		newImageStream, err = r.imageStreamRegistry.UpdateImageStream(ctx, imageStream, false, &metav1.UpdateOptions{})
 	}
 	if err != nil {
-		return nil, false, err
+		// return true for canRetry if we had a failure for resource versions
+		// and no resource version was passed in
+		return nil, false, err, canRetry
```
deads2k, line 342: This is the only point where a retry is suitable and it is *only* suitable when the error is a conflict error.  BeforeUpdate and updateValidation should be fatal.

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -206,20 +207,44 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
+		if canRetry {
+			return updateErr
+		}
+
+		// if not canRetry return nill for the error, we captured it in updateErr and will pass it back up.
+		return nil
+	})
+
 	if err != nil {
 		return nil, false, err
```
deads2k, line 227: is it more appropriate to prefer the updateError if it is not nil in this case?

pkg/image/apiserver/registry/imagestreamtag/rest.go
neisw: updated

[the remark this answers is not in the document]
neisw: updated comments

pkg/image/apiserver/registry/imagestreamtag/rest.go
neisw: makes sense

pkg/image/apiserver/registry/imagestreamtag/rest.go
neisw: removed the others

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -206,20 +207,54 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
+		}
+
+		// if not canRetry return nill for the error, we captured it in updateErr and will pass it back up.
+		return nil
+	})
+
+	// prefer updateErr over err
+	// but if either are non nil return error
+	if updateErr != nil {
+		err = updateErr
```
deads2k, line 229: can we just `return nil, false, updateErr`?

deads2k:
one remaining comment and this code looks correct to me.

pkg/image/apiserver/registry/imagestreamtag/rest.go
neisw: yep

pkg/image/apiserver/registry/imagestreamtag/imagestreamregistrytester.go
```diff
@@ -0,0 +1,131 @@
...
+	"context"
+
+	imageapi "github.com/openshift/openshift-apiserver/pkg/image/apis/image"
+	"github.com/openshift/openshift-apiserver/pkg/image/apiserver/registry/imagestream"
+	metainternal "k8s.io/apimachinery/pkg/apis/meta/internalversion"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+type ImageStreamRegistryTester struct {
```
deads2k, line 13: If this is only used for tests, please name the file with the  _test suffix.

deads2k, line 13: lowercase, no need to export, right?

pkg/image/apiserver/registry/imagestreamtag/imagestreamregistrytester.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// which apis are we providing responses for
+	registryApiTesters map[string]*ApiTester
+}
+
+func NewImageStreamRegistryTester(registry imagestream.Registry, apiTesters map[string]*ApiTester) *ImageStreamRegistryTester {
+	tester := &ImageStreamRegistryTester{}
+
+	tester.Registry = registry
+	tester.registryApiTesters = apiTesters
```
aojea: do you want to sanity check that this is not nil? 
and initialize the map if is nil?

neisw: Yes, thank you.  Added a check in the extract api as well just to be sure.

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,576 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+				},
+				Image: imageapi.Image{ObjectMeta: metav1.ObjectMeta{Name: "10"}, DockerImageReference: "foo/bar/baz"},
+			},
+			expectNilResult: true,
+			expectError:     true,
+		},
+	}
+
+	for name, tc := range tests {
+		func() {
```
aojea, line 711:
```suggestion
			t.Run(name, func(t *testing.T) {
```

pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
```diff
@@ -0,0 +1,138 @@
...
+func NewApiTester() *ApiTester {
+	tester := &ApiTester{}
+	tester.callCount = 0
+	tester.callResponses = make(map[int]ApiResponse)
+
+	return tester
+}
+
+func (a *ApiTester) callComplete() {
+	a.callCount++
```
aojea, line 59: do you need this counter to be thread safe?

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,576 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+				t.Errorf("%s: obj is not an ImageStreamTag: %#v", name, tc.istag)
+				return
+			}
+
+			result, create, err := storage.Update(ctx, istag.Name, rest.DefaultUpdatedObjectInfo(istag), rest.ValidateAllObjectFunc, rest.ValidateAllObjectUpdateFunc, false, &metav1.UpdateOptions{})
+
+			gotErr := err != nil
+			if e, a := tc.expectError, gotErr; e != a {
+				t.Errorf("%s: Expected err=%v: got %v: %v", name, e, a, err)
+				return
```
aojea, line 805: if you run this inside `t.Run` you can use `t.Fatalf()` directly

aojea: this applies to all t.Errorf and returns

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,576 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+			}
+			if tc.expectError {
+
+				if !gotErr {
+					t.Errorf("%s: Expected err but did not receive one", name)
+					return
+				}
```
aojea, line 812: this should be asserted in https://github.com/openshift/openshift-apiserver/pull/304/files#r934694000

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,576 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+}
+
+func TestUpdateRetryImageStreamTag(t *testing.T) {
+	tests := map[string]struct {
+		istag                     runtime.Object
+		expectError               bool
+		createUpdateConflictError bool
+		createUpdateInvalidError  bool
+		createCreateConflictError bool
+		createCreateInvalidError  bool
```
aojea, line 866: these fields are really indicating different test cases, instead of adding conditions inside the test you should check how to inject those behaviours from the testcase params, or just use independent tests ...

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,576 @@ func TestCreateImageStreamTag(t *testing.T) {
 		}()
 	}
 }
+
+func TestUpdateImageStreamTag(t *testing.T) {
+	tests := map[string]struct {
+		istag                     runtime.Object
+		expectError               bool
+		createUpdateConflictError bool
```
aojea, line 582: https://github.com/openshift/openshift-apiserver/pull/304/files#r934702928

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -206,20 +207,54 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 }
 
 func (r *REST) Update(ctx context.Context, tagName string, objInfo rest.UpdatedObjectInfo, createValidation rest.ValidateObjectFunc, updateValidation rest.ValidateObjectUpdateFunc, forceAllowCreate bool, options *metav1.UpdateOptions) (runtime.Object, bool, error) {
-	name, tag, err := nameAndTag(tagName)
+	var result runtime.Object
+	var created bool
+	var updateErr error
+
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
aojea, line 214: we somehow have to add a comment with all these things David explained in slack  https://coreos.slack.com/archives/CB48XQ4KZ/p1659371510569929?thread_ts=1659361243.293439&cid=CB48XQ4KZ

> > It's an unconditional imagestreamtag update, but a conditional update on the imagestream
> 
> so the imagestreamtag is always updated. But the imagestream update that is done will not backlevel say an annotation update by doing an unconditional update
> 

or nobody will understand this in the future :smile:

aojea: ah, you already did, in the `update` method

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
neisw: thanks, all set

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,576 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+
+			if !ok {
+				t.Errorf("%s: obj is not an ImageStreamTag: %#v", name, tc.istag)
+				return
+			}
+
+			result, create, err := storage.Update(ctx, istag.Name, rest.DefaultUpdatedObjectInfo(istag), rest.ValidateAllObjectFunc, rest.ValidateAllObjectUpdateFunc, false, &metav1.UpdateOptions{})
+
+			gotErr := err != nil
+			if e, a := tc.expectError, gotErr; e != a {
```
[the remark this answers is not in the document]
aojea: it is a nit, choose whatever you prefer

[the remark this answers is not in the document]
neisw: The two tests I added were cloned from existing but I am open to changing.

pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go
neisw: definitely safer to do so, thanks

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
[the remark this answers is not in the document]
neisw: It's good feedback, all of it.  I think I was able to address it all with the latest update.

pkg/image/apiserver/registry/imagestreamtag/rest.go
neisw: I can add more if what is there isn't a complete summarization of David's explanation.

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
neisw: Yes, it was starting to feel like too many flags.  Thanks for the feedback.

pkg/image/apiserver/registry/imagestreamtag/rest.go
aojea: add a comment on top of the retryOnConflict, so it is clear why do we have this retry here

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -568,3 +574,728 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+					apiTesters := make(map[string]*ApiTester)
+
+					if tc.stagedError != nil {
+
+						apiTester := NewApiTester()
+						updateResponses := make(map[int32]ApiResponse)
+						apiResponse := NewApiResponse()
+
+						apiResponse.response["error"] = tc.stagedError
+						updateResponses[0] = apiResponse
```
aojea, line 684: if only has a response for the first call? ... how it works for the retry? doesn't receive two requests?

aojea: I find hard to understand how you are mocking the imagestream registry

neisw: https://github.com/neisw/openshift-apiserver/blob/f9c54737775514a5be102e841c9cde7eb453d70d/pkg/image/apiserver/registry/imagestreamtag/imagestreamregistry_test.go#L85-L94

If there is no response extracted (the second call will have the callCount == 1) then it passes through to the main imagestream.Registry

neisw: Pushed a new test with multiple failures, see https://github.com/neisw/openshift-apiserver/blob/51fb374b7ac39278b4da148061cbb5c94b4a6b6a/pkg/image/apiserver/registry/imagestreamtag/rest_test.go#L811-L814

deads2k:
from the job

> Running `gofmt -s -l` on 468 file(s).
./pkg/image/apiserver/registry/imagestreamtag/rest_test.go
verify-gofmt failed - please run `make update-gofmt`

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
aojea: thanks for explaining

aojea:
lgtm

deads2k:
> lgtm

/lgtm
/approve

PR #290, opened 2022-06-29
IR-258: ImageStreamImport manifest list support

flavianmissi:

PR #309, opened 2022-09-19
IR-269: Support multi-arch images in ImageStreamLayers

dmage:


dmage:
FTR - it'll be rebased once #290 is merged.

pkg/image/apiserver/registry/imagestream/etcd/image.go
```diff
@@ -35,30 +35,47 @@ type ImageListWatch interface {
 	Watch(context.Context, metav1.ListOptions) (watch.Interface, error)
 }
 
-type imageLayerIndex struct {
-	informer cache.SharedIndexInformer
+type MockImageLayerIndex struct {
```
flavianmissi, line 38: nit: do you think this belongs in a test or test util file? feels odd to have a mock in production code like this.
this is a nit because I do realize this was already that way, you just renamed it.

flavianmissi (commented):
left some comments - lgtm overall

pkg/image/apiserver/registry/imagestream/etcd/etcd.go
```diff
@@ -210,6 +210,61 @@ func (r *LayersREST) Get(ctx context.Context, name string, options *metav1.GetOp
...
+		if _, ok := isl.Blobs[layer.Name]; !ok {
+			isl.Blobs[layer.Name] = imageapi.ImageLayerData{LayerSize: &layer.LayerSize, MediaType: layer.MediaType}
+		}
+	}
+
+	if blob := entry.Config; blob != nil {
+		reference.Config = &blob.Name
+		if _, ok := isl.Blobs[blob.Name]; !ok {
+			if blob.LayerSize == 0 {
+				// only send media type since we don't the size of the manifest
```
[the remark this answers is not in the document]
dmage: I moved it as is, but yeah, I'll fix it

pkg/image/apiserver/registry/imagestream/etcd/etcd.go
```diff
@@ -210,6 +210,61 @@ func (r *LayersREST) Get(ctx context.Context, name string, options *metav1.GetOp
...
+				isl.Blobs[blob.Name] = imageapi.ImageLayerData{LayerSize: &blob.LayerSize, MediaType: blob.MediaType}
+			}
+		}
+	}
+
+	for _, manifest := range entry.Manifests {
+		reference.Manifests = append(reference.Manifests, manifest.Digest)
+	}
+
+	// the image manifest is always a blob - schema2 images also have a config blob referenced from the manifest
```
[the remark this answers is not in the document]
dmage: no, I moved it as is; it doesn't look correct to me, but perhaps the very low-level image registry internals (that indeed don't distinguish blobs and manifests) rely on this

pkg/image/apiserver/registry/imagestream/etcd/etcd.go
```diff
@@ -210,6 +210,61 @@ func (r *LayersREST) Get(ctx context.Context, name string, options *metav1.GetOp
...
+		reference.Layers = append(reference.Layers, layer.Name)
+		if _, ok := isl.Blobs[layer.Name]; !ok {
+			isl.Blobs[layer.Name] = imageapi.ImageLayerData{LayerSize: &layer.LayerSize, MediaType: layer.MediaType}
+		}
+	}
+
+	if blob := entry.Config; blob != nil {
+		reference.Config = &blob.Name
+		if _, ok := isl.Blobs[blob.Name]; !ok {
+			if blob.LayerSize == 0 {
```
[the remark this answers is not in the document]
dmage: I'd prefer to extract this code into a new function without significant modifications. But it seems to be ok, we have metadata for layers, but we don't have metadata for image configs. So this condition will almost always be true, unless there is a strange image that uses an image config as its layer.

pkg/image/apiserver/registry/imagestream/etcd/image.go
dmage: that's a tricky one, other package doesn't really need this ImageLayerIndex, they all discard `LayersREST` from `imagestreametcd.NewRESTWithLimitVerifier`. So they all can use `nil` instead of `MockImageLayerIndex`. Are we ok with `nil` as an input?

pkg/image/apiserver/registry/imagestream/etcd/etcd.go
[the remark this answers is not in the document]
flavianmissi: oh I missed that, let's leave it be then

pkg/image/apiserver/registry/imagestream/etcd/image.go
flavianmissi: hmm I'd rather avoid using `nil` as an input if we can. I guess the option left is to move it to a test util package? could https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/testutil/util.go be a good place for it?
as I said this is a nit, we can also look into as a follow-up if it's something we want to pursue.

pkg/image/apis/image/validation/validation.go
```diff
@@ -801,3 +801,21 @@ func getWhitelistTransportForFlag(insecure, allowSecureFallback bool) whitelist.
 	}
 	return whitelist.WhitelistTransportSecure
 }
+
+// ValidateImageStreamLayers ensures that given object is valid.
+func ValidateImageStreamLayers(isl *imageapi.ImageStreamLayers) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for name, ibr := range isl.Images {
```
deads2k, line 808: Per the API docs, if len(manifests) >0, then layers and configs must be len=0.  This should be enforced here.

pkg/image/apis/image/validation/validation.go
```diff
@@ -801,3 +801,21 @@ func getWhitelistTransportForFlag(insecure, allowSecureFallback bool) whitelist.
 	}
 	return whitelist.WhitelistTransportSecure
 }
+
+// ValidateImageStreamLayers ensures that given object is valid.
+func ValidateImageStreamLayers(isl *imageapi.ImageStreamLayers) field.ErrorList {
```
deads2k, line 806: the validation when `ImageMissing` is true is also missing.

dmage: Added validation for when ImageMissing is true.

pkg/image/apis/image/validation/validation.go
dmage: Done.

xiuwang:
@dmage This pr need rebase to resolve conflicts

PR #331, opened 2022-11-14
move the deployer role and binding to the ocm-o

deads2k:
after https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/270 merges, this will remove the old so we don't end up with conflict content.


soltysh (approved):
/lgtm
/hold
for https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/270

PR #344, opened 2023-01-03
IR-269: Bump openshift/api

dmage:
Bumping openshift/api and updating all generated files as a preparation step for manifest lists support in the layers API. The final implementation will be completed by #309.


dmage:
/assign @sanchezl
This is just bumping openshift/api for easier review.

PR #346, opened 2023-01-07
Adds Route API to set/delete headers.

miheer:


miheer:
new PR https://github.com/openshift/openshift-apiserver/pull/374

PR #347, opened 2023-01-13
[IR-342] Add OS/arch information to image stream status

werled:
https://issues.redhat.com/browse/IR-342

Includes the changes proposed in openshift/api#1392 and should only be merged if those changes land.


werled:
I'm looking into to the failed unit tests, which I suspect are due to either the fuzzer not being aware of the changes or the dependencies being bumped for the PR.

There are currently concerns wrt. to increased object size (see https://issues.redhat.com/browse/IR-342), which have to be addressed before proceeding with this PR.

/hold

PR #349, opened 2023-01-25
IR-270: support creation of image objects representing manifest lists

flavianmissi:


pkg/image/apiserver/registry/image/strategy.go
```diff
@@ -50,6 +51,31 @@ func (s imageStrategy) PrepareForCreate(ctx context.Context, obj runtime.Object)
 	if err := internalimageutil.InternalImageWithMetadata(newImage); err != nil {
 		utilruntime.HandleError(fmt.Errorf("Unable to update image metadata for %q: %v", newImage.Name, err))
 	}
+
+	manifest, _, err := distribution.UnmarshalManifest(
+		newImage.DockerImageManifestMediaType,
+		[]byte(newImage.DockerImageManifest),
+	)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("Unable to parse manifest for %q: %v", newImage.Name, err))
```
dmage, line 60: is it ok to continue after this error?
errors should start with a lower case letter.

flavianmissi: yeah there should be no issues continuing. the casing is fixed.

PR #352, opened 2023-02-17
OCPBUGS-7689: Fix a project validation error due to empty string value

dinhxuanvu:
When a new label with non-empty value is added to a project, validation error will occur due to missing label is mistaken as a label with empty string value.

The fix is to verify if the label exists before comparing its value.

Signed-off-by: Vu Dinh <vudinh@outlook.com>

### DIFF
--- a/pkg/project/apis/project/validation/validation.go
+++ b/pkg/project/apis/project/validation/validation.go
@@ -80,9 +80,10 @@ func ValidateProjectUpdate(newProject *projectapi.Project, oldProject *projectap
 		}
 	}
 
+	// Check to ensure no new labels are added or existing one being deleted or modified
 	for name, value := range newProject.Labels {
-		if value != oldProject.Labels[name] {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "labels").Key(name), value, "field is immutable, , try updating the namespace"))
+		if oldValue, ok := oldProject.Labels[name]; !ok || oldValue != value {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "labels").Key(name), value, "field is immutable, try updating the namespace"))
 		}
 	}
 	for name, value := range oldProject.Labels {
--- a/pkg/project/apis/project/validation/validation_test.go
+++ b/pkg/project/apis/project/validation/validation_test.go
@@ -314,7 +314,7 @@ func TestValidateProjectUpdate(t *testing.T) {
 			T: field.ErrorTypeInvalid,
 			F: "metadata.annotations[openshift.io/node-selector]",
 		},
-		"updating label": {
+		"modifying label": {
 			A: projectapi.Project{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:            "project-name",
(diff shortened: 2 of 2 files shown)


dinhxuanvu:
@tkashem PTAL

pkg/project/apis/project/validation/validation.go
```diff
@@ -81,8 +81,10 @@ func ValidateProjectUpdate(newProject *projectapi.Project, oldProject *projectap
 	}
 
 	for name, value := range newProject.Labels {
-		if value != oldProject.Labels[name] {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "labels").Key(name), value, "field is immutable, , try updating the namespace"))
+		if oldValue, ok := oldProject.Labels[name]; ok {
```
sttts, line 84: would have expected the inverse: be more strict aka `if ....; !ok || oldValue != value { error }`

pkg/project/apis/project/validation/validation_test.go
```diff
@@ -250,6 +250,24 @@ func TestValidateProjectUpdate(t *testing.T) {
...
+			Annotations: map[string]string{
+				oapi.OpenShiftDescription:      "This is a description",
+				oapi.OpenShiftDisplayName:      "display name change",
+				projectapi.ProjectNodeSelector: "infra=true, env = test",
+			},
+			Labels: map[string]string{"label-name": "value", "label-name-2": "value2"},
+		},
+	}
+
+	errs = ValidateProjectUpdate(addLabel, project)
```
tkashem, line 266: we also want to simulate adding a new label with empty string, right? 
```
objWithNewLabel := project.DeepCopy()
objWithNewLabel.Labels["new-label"]=""

errs = ValidateProjectUpdate(objWithNewLabel, project)
```

pkg/project/apis/project/validation/validation.go
dinhxuanvu: Hi Stefan. I'm a bit confused by your suggestion. Why would we error out when a new label (when `ok = false`), which doesn't exist yet, being added?

tkashem: judging from the error, i think it's because the labels are immutable once created, but @sttts can confirm

dinhxuanvu: The labels are immutable after creation but there is nothing wrong with adding a new label or at least I don't see anything in the current validation code indicates that.

dinhxuanvu: This issue is now resolve as I checked with Maciej and it seems labels are not immutable at all including adding, deleting and modifying.

dinhxuanvu:
@tkashem PTAL

pkg/project/apis/project/validation/validation_test.go
```diff
@@ -326,6 +326,18 @@ func TestValidateProjectUpdate(t *testing.T) {
 			T: field.ErrorTypeInvalid,
 			F: "metadata.labels[label-name]",
 		},
+		"adding label": {
```
tkashem, line 329: i don't see a case for `adding new label witn non empty values`, can we add it?

dinhxuanvu: Updated.

tkashem:
/lgtm
/approve

(I believe this is backport worthy)

PR #366, opened 2023-05-02
WRKLDS-728: Disable apiservers

ingvagabund:
Extend the OA init code to explicitly disable individual apiservers. The change is important for explicitly disabling DeploymentConfigs and Builds API through capabilities. The controlplane configuration on-disk data type gets rendered by OA operator in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/532.

### DIFF
--- a/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go
+++ b/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go
@@ -14,6 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	genericmux "k8s.io/apiserver/pkg/server/mux"
@@ -57,6 +58,11 @@ import (
 	_ "github.com/openshift/openshift-apiserver/pkg/api/install"
 )
 
+var configurableAPIList = sets.New[openshiftcontrolplanev1.OpenShiftAPIserverName](
+	openshiftcontrolplanev1.OpenShiftAppsAPIserver,
+	openshiftcontrolplanev1.OpenShiftBuildAPIserver,
+)
+
 type OpenshiftAPIExtraConfig struct {
 	// we phrase it like this so we can build the post-start-hook, but no one can take more indirect dependencies on informers
 	InformerStart func(stopCh <-chan struct{})
@@ -89,6 +95,9 @@ type OpenshiftAPIExtraConfig struct {
 	RESTMapper                *restmapper.DeferredDiscoveryRESTMapper
 
 	ClusterQuotaMappingController *clusterquotamapping.ClusterQuotaMappingController
+
+	// apiServers holds information about enabled/disabled API servers
+	APIServers openshiftcontrolplanev1.APIServers
 }
 
 // Validate helps ensure that we build this config correctly, because there are lots of bits to remember for now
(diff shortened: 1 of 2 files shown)


dgrisonnet:
I think we will need to touch some other parts of the codebase to truly be able to enable/disable APIs. The current change only stop serving the APIs, but the underneath logic will still be executed even when the APIs are disabled.

The first one that I can see is the post start hooks: https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go#L444-L473. Some of these will not be needed if their API is disabled.

Another part that will need to be gated on the config are the various caches configurations in https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/config.go. I don't think we should create the cache of an API that is disabled. Same goes for the controllers that are normally used to fill these cache and respond to the apiserver.

There might be some other parts of the code that will need to be tweaked, but I think we should also try to enable/disable the logic behind the APIs.

dgrisonnet:
I like the change to have an allow list with the API that can be disabled today. That said we still have the same problem as before from https://github.com/openshift/openshift-apiserver/pull/366#issuecomment-1638153133 where the inner logic (controller, cache, ...) will still be executed even if an API is disabled no?

Also, I think it would be a great addition to add a log line and maybe even a metric telling which API is enabled and which one is disabled.

ingvagabund:
> I like the change to have an allow list with the API that can be disabled today. That said we still have the same problem as before from https://github.com/openshift/openshift-apiserver/pull/366#issuecomment-1638153133 where the inner logic (controller, cache, ...) will still be executed even if an API is disabled no?

Checking the PostStart hooks I don't see anything related to DCs/Builds under https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go#L453-L484. There's also no DCs/Builds informer mentioned in https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go#L98-L142 or https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/informers.go#L90-L99.

dgrisonnet:
This looks good to me overall, but we might also want to disable these admission plugins https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftadmission/register.go#L37-L38 no?

ingvagabund:
Those are registering handlers, which should not get invoked at all. Plus, `RegisterOpenshiftAdmissionPlugins` is invoked in `RegisterAllAdmissionPlugins` which is invoked in `init()` function. Too late to read the config and disable the handler.

dgrisonnet:
We could maybe leverage the admission options to disable them? https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/config.go#L212

But as you mentioned it should be impossible to invoke them so disabling them doesn't bring much value

pkg/cmd/openshift-apiserver/openshiftapiserver/openshift_apiserver.go
```diff
@@ -416,15 +425,59 @@ func addAPIServerOrDie(delegateAPIServer genericapiserver.DelegationTarget, apiS
...
+	// All API servers are enabled by default (nothing new to enable -> ignore the list of enabled versions)
+	for name, initFnc := range apiServerInitializers {
+		if _, exists := apiServers[name]; exists {
+			// All API servers are serving v1 resources
+			if sets.NewString(apiServers[name].DisabledVersions...).Has("v1") {
+				continue
+			}
+		}
+		delegateAPIServer = addAPIServerOrDie(delegateAPIServer, initFnc)
+	}
```
dgrisonnet, line 480: nit: could you move that code inside a function to make `New` more readable?

dgrisonnet (approved):
/lgtm
/approve

PR #367, opened 2023-05-03
[WIP] Remove system:unauthenticated apiGroup for system:webhook ClusterRole

mbharatk:
Jira: https://issues.redhat.com/browse/OCPBUGS-12365

Removing the system:unauthenticated apiGroup in system:webhook ClusterRoleBinding as the system:unauthenticated user group allows unauthenticated users to get and create buildconfigs/webhooks


coreydaley:
@mbharatk It looks like there is a unit test that needs updated, but other than that everything else passed, weird. :)

mbharatk:
@coreydaley I have tried to find out which test to update but I only found https://github.com/openshift/openshift-apiserver/blob/master/pkg/bootstrappolicy/policy_test.go, not sure if this is correct. Could you point me to where I should be looking?

PR #382, opened 2023-07-06
add featuregates to server

deads2k:


JoelSpeed:
Given the enhancement for route external certificates has not yet merged, I'm thinking we should probably revert the API change until the enhancement has merged, in which case, we won't need this PR yet right?

deads2k:
> Given the enhancement for route external certificates has not yet merged, I'm thinking we should probably revert the API change until the enhancement has merged, in which case, we won't need this PR yet right?

I'd keep the gate and add the field stripping.

deads2k:
Now with field stripping.

pkg/api/validation/register.go
```diff
@@ -60,7 +61,7 @@ func registerAll() {
 	Validator.MustRegister(&projectapi.Project{}, false, projectvalidation.ValidateProject, projectvalidation.ValidateProjectUpdate)
 	Validator.MustRegister(&projectapi.ProjectRequest{}, false, projectvalidation.ValidateProjectRequest, nil)
 
-	Validator.MustRegister(&routeapi.Route{}, true, routevalidation.ValidateRoute, routevalidation.ValidateRouteUpdate)
+	Validator.MustRegister(&routeapi.Route{}, true, routeOptionsAdapter, routeUpdateOptionsAdapter)
```
p0lyn0mial, line 64: Why does it have to be registered here? Where is it being used?

pkg/cmd/openshift-apiserver/cmd.go
```diff
@@ -101,6 +101,10 @@ func (o *OpenShiftAPIServer) Validate() error {
 
 // RunAPIServer takes the options, starts the API server and waits until stopCh is closed or initial listening fails.
 func (o *OpenShiftAPIServer) RunAPIServer(stopCh <-chan struct{}) error {
+	if err := features.InitializeFeatureGates(feature.DefaultMutableFeatureGate, configv1.FeatureGateRouteExternalCertificate); err != nil {
```
p0lyn0mial, line 104: Why is it initialized in this place? What happens if I disable the feature at the command-line argument level while it's enabled by default? It seems to me that in such a case, it will be overwritten and enabled. Is that correct?

pkg/route/apis/route/validation/validation.go
```diff
@@ -17,17 +17,22 @@ func toRouteV1(internal *routeapi.Route) (*routev1.Route, field.ErrorList) {
...
 }
 
+// move to library-go for plumbing
+type RouteValidationOptions struct {
+	AllowExternalCertificates bool
+}
+
 // ValidateRoute tests if required fields in the route are set.
-func ValidateRoute(route *routeapi.Route) field.ErrorList {
+func ValidateRoute(route *routeapi.Route, opts RouteValidationOptions) field.ErrorList {
```
p0lyn0mial, line 26: I haven't seen a unit test for this method. Worth adding ?

pkg/cmd/openshift-apiserver/openshiftapiserver/config.go
```diff
@@ -111,6 +112,15 @@ func NewOpenshiftAPIConfig(config *openshiftcontrolplanev1.OpenShiftAPIServerCon
 	// non-root apiservers must set this value to false
 	genericConfig.Config.SkipOpenAPIInstallation = true
 
+	// set the global featuregates.`
+	warnings, err := features.SetFeatureGates(config.APIServerArguments, feature.DefaultMutableFeatureGate)
+	if err != nil {
+		return nil, err
+	}
+	for _, warning := range warnings {
+		klog.Warning(warning)
```
p0lyn0mial, line 121: I think that in practice this will warn about ignored unknown flags.

pkg/route/apiserver/registry/route/strategy.go
```diff
@@ -61,6 +61,10 @@ func (s routeStrategy) PrepareForCreate(ctx context.Context, obj runtime.Object)
 	route := obj.(*routeapi.Route)
 	route.Status = routeapi.RouteStatus{}
 	stripEmptyDestinationCACertificate(route)
+
+	if !s.allowExternalCertificates && route.Spec.TLS != nil {
+		route.Spec.TLS.ExternalCertificate.Name = ""
```
p0lyn0mial: why not to return an error ?

pkg/route/apiserver/registry/route/strategy.go
```diff
@@ -74,6 +78,10 @@ func (s routeStrategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Ob
 	if len(route.Spec.Host) == 0 {
 		route.Spec.Host = oldRoute.Spec.Host
 	}
+
+	if !s.allowExternalCertificates && route.Spec.TLS != nil {
+		route.Spec.TLS.ExternalCertificate.Name = ""
```
p0lyn0mial: won't this cause some disruption ? It seems like previously admitted routes will get this field cleared depending on the `allowExternalCertificates` feature ?

will there be a controller for updating `routes` ?

p0lyn0mial: how are we going to ensure consistency? (all obj will be cleared after turning off the FG)

p0lyn0mial: won't this be checked in the `ValidateRoute` method?

pkg/route/apis/route/validation/validation.go
```diff
@@ -17,17 +17,22 @@ func toRouteV1(internal *routeapi.Route) (*routev1.Route, field.ErrorList) {
 	return &external, nil
 }
 
+// move to library-go for plumbing
```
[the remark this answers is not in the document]
deads2k: > is this a `TODO` ?

network edge team owes this and they'll plug it in here.

pkg/route/apis/route/validation/validation.go
```diff
@@ -17,17 +17,22 @@ func toRouteV1(internal *routeapi.Route) (*routev1.Route, field.ErrorList) {
...
 // ValidateRoute tests if required fields in the route are set.
-func ValidateRoute(route *routeapi.Route) field.ErrorList {
+func ValidateRoute(route *routeapi.Route, opts RouteValidationOptions) field.ErrorList {
 	external, errs := toRouteV1(route)
 	if len(errs) > 0 {
 		return errs
 	}
 
-	return routevalidation.ValidateRoute(external)
+	return routevalidation.ValidateRoute(external /*, opts*/)
```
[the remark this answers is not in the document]
deads2k: > why not to change the library-go first?

Leaving this for network edge.  This PR is about demonstrating where the plumbing lies.

pkg/route/apis/route/validation/validation.go
deads2k: > I haven't seen a unit test for this method. Worth adding ?

Ben moved it to library-go a while back.

pkg/route/apiserver/registry/route/strategy.go
deads2k: > why not to return an error ?

Because the way kube APIs work, if newer manifests are used against older servers, the unknown fields are stripped (since before v1).  We could change the semantic and there are some valid reasons to do so (the object you wanted wasn't created), but it is lossy since the skew is still only at the point where the type can be deserialized.  This is backwards compatible with 4.13 behavior for instance.

pkg/cmd/openshift-apiserver/cmd.go
deads2k: The initialization here ensures that all featuregates used by this binary are registered and have default values so that openshift-apiserver (this binary) can interrogate them later.  The call to `features.SetFeatureGates` (called *after* this line) will set the featuregate enable/disable values based on the CLI flag.

pkg/route/apiserver/registry/route/strategy.go
p0lyn0mial: I understand that the old kas will drop the field during decoding. 
I also understand that we are introducing a new field to the existing type, and we want to have it hidden behind a feature gate (FG).

What I don't understand is:
- Why don't we stop accepting data into this field until the FG is enabled?
- Why are we dropping data on an update? (Such a solution seems counterintuitive, and I think it's not what users expect)
 
In my opinion, what we should do is:
- Don't accept (return an error) data into this field unless the FG is enabled.
- If some objects with the new field set already exist in etcd, we should allow reading and writing them (update) without dropping the data (after the FG was on).

I believe this approach would be acceptable because, after the first release, all servers will have this new field (allowing roundtrip etcd data), but during validation, we won't allow data into this field until the FG feature is enabled (which occurs after a cluster upgrade)

pkg/cmd/openshift-apiserver/cmd.go
p0lyn0mial: Ok. Thanks. I just wanted to make sure that the initialisation happens before we read the CLI flags.
It wasn't clear to me when I was reading the code.

pkg/api/validation/register.go
p0lyn0mial: Ok. I think I found the test.

https://github.com/openshift/openshift-apiserver/blob/12686d8a51ec83b3cba05c28d341593e008162b0/pkg/api/validation/validation_test.go#L41

pkg/route/apis/route/validation/validation.go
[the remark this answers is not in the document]
p0lyn0mial: Please add an explicit `TODO` for the network team.
It would be perfect if you could also include a link to some issue.

pkg/route/apis/route/validation/validation.go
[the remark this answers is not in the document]
p0lyn0mial: Please add an explicit `TODO` for the network team and a link to some issue.

pkg/cmd/openshift-apiserver/openshiftapiserver/config.go
p0lyn0mial: I don't know, maybe it would be better to simply stop the program?

pkg/route/apiserver/registry/route/strategy.go
p0lyn0mial: I discussed this point with David in Slack. 
Techpreview feature gates are one-way switch.
We also don't allow turning it back off and we don't allow upgrades.
It is also not possible to drop fields on read because we don't have an injection point for that.

pkg/route/apis/route/validation/validation.go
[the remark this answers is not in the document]
p0lyn0mial: David promised to address it in a follow-up PR.

PR #385, opened 2023-07-31
CFE-885:Feature route external certificate validation

thejasn:
Description
---
Updates route validations based on https://github.com/openshift/enhancements/pull/1307 which introduces a new field in the route API `externalCertificate` behind the TP feature gate. 

### Changes
- `ValidateRoute()` has been updated to pass additional args (secrets lister and subjectaccessreviewer) to validate 
  `externalCertificate` and the rbac required.
- `ValidateHostUpdate()` has also been updated since updating route host/subdomain is also affected by updating 
  certificates on the route. 
- `ValidateHostExternalCertificate()` has been introduced as part of the validations done during route updates, this function specifically checks if the user has the correct permissions on the `custom-host` sub-resource. Additional details can be found on the EP.

Follow up from: https://github.com/openshift/library-go/pull/1549


chiragkyal:
/close 
Being continued in https://github.com/openshift/openshift-apiserver/pull/407

PR #386, opened 2023-08-17
OCPBUGS-16514: use import mode to manifest cache key

flavianmissi:
When trying to import two images with the same digest but different import mode, the code will currently bug out and import both images as Legacy. Using the import mode in the manifest cache key solves the problem.


flavianmissi:
I'm not sure if we should rush this into master before 4.15 branching.
/hold

PR #387, opened 2023-09-04
OCPBUGS-17674: pkg/image: avoid unnecessary service lookups when registry is removed

flavianmissi:
also remove the now obsolete ip resolution code.

we have received reports of the openshift-apiserver becoming unstable due to repeated requests to /api/v1/namespaces/default/services/docker-registry when the internal registry is removed, caused by BuildSpecs not using DockerImage as build output.


dgrisonnet (approved):
/lgtm
/approve

flavianmissi:
first attempt failed to clone the bug in jira.
/cherry-pick release-4.14

PR #388, opened 2023-09-05
OCPBUGS-5823: remove unused service-ca roles

stlaz:


dgrisonnet (commented):
This change need to be done over two releases in order to first remove the resources from the existing clusters and then remove the manifests from the repo.

To remove them from the clusters, you need to add the `release.openshift.io/delete: "true"` annotation to the manifests to tell CVO to cleanup the resource.

dgrisonnet (approved):
/approve

PR #419, opened 2024-03-06
OCPBUGS-30279: vendor upgrade runtime-utils

QiWang19:


QiWang19:
@flavianmissi PTAL

flavianmissi:
Thanks for this @QiWang19! Do you think this is the problem @hongkailiu was facing in relation to OCPBUGS-30279? I haven't been able to confirm that myself yet.

flavianmissi:
I can confirm that it works now on GCP (`launch openshift/openshift-apiserver#419,4.16.0-ec.3 gcp`) - I followed the steps described in the bug comments and no longer get an error. 
cc @hongkailiu

sanchezl (approved):
/lgtm

QiWang19:
@flavianmissi I think we can backport this to 4.15 and 4.14? OCP has backported to these version to support both ICSP, ITMS resources. Although the imagestream uses its own copy of registries config struct, we can backport for some kind of compatibility?

flavianmissi:
Backporting should be fine, but we need to do it manually so that we vendor commits from the correct branches - if we just use the cherry pick bot we'll endup vendoring commits from 4.16 on the 4.15 branch and so on (for example: `v3.0.0-20240215131201-6b2f5d2f1f43` is a commit from the master branch of the docker-distribution repo and although the same commit is present in 4.15 and 4.14 branches it has different commit hashes).
Another thing is that release-4.14 branch of openshift/runtime-utils does not contain the commit allowing ICSP and IDMS objects to coexist ([release-4.14 commits](https://github.com/openshift/runtime-utils/commits/release-4.14) vs [release-4.15 commits](https://github.com/openshift/runtime-utils/commits/release-4.15)), so unless that commit is also backported to runtime-utils 4.14 branch we can only backport this bump down to 4.15 - your call.

QiWang19:
@flavianmissi FYI, this runtime-utils PR is up for 4.14 backport https://github.com/openshift/runtime-utils/pull/21

PR #420, opened 2024-03-13
NO-ISSUE: go mod tidy && go mod vendor

p0lyn0mial:
pulls https://github.com/openshift/kubernetes-apiserver/pull/55 

doesn't require any changes because the config is created "manually" - https://github.com/openshift/openshift-apiserver/blob/master/pkg/cmd/openshift-apiserver/openshiftapiserver/config.go#L58 and the apf is never enabled (even if the flag says so)


vrutkovs (approved):
/lgtm

PR #428, opened 2024-04-18
OCPBUGS-32444: Bump golang/x/net to v0.23.0

tkashem:


p0lyn0mial:
/hold

I think that we should bump https://github.com/openshift/kubernetes-apiserver/blob/openshift-apiserver-4.16-kubernetes-1.29.2/go.mod#L36 instead. Then update `kubernetes-apiserver` used by this binary.

p0lyn0mial:
> I think that we should bump https://github.com/openshift/kubernetes-apiserver/blob/openshift-apiserver-4.16-kubernetes-1.29.2/go.mod#L36 instead. Then update kubernetes-apiserver dep used by this binary.

According to https://go.dev/ref/mod#minimal-version-selection it will work.

/hold cancel

PR #435, opened 2024-05-13
OCPBUGS-33453: pkg/bootstrappolicy: support anonymous access to self access reviews

flavianmissi:
/assign @ibihim


flavianmissi:
It looks like the right way to fix anonymous pulls support for the image registry is to change an existing CR - I'm syncing with the AUTH team to figure out the exact steps. this means this change is unnecessary.
/close

PR #439, opened 2024-07-08
[WIP] OCPBUGS-35036: fix panic from image importer

tkashem:


pkg/image/apiserver/importer/importer.go
```diff
@@ -230,13 +231,15 @@ func (imp *ImageStreamImporter) importImages(ctx context.Context, isi *imageapi.
...
 			tags[tag] = append(tags[tag], i)
 			if len(tags[tag]) == 1 {
-				repo.Tags = append(repo.Tags, importTag{
-					Name:       defaultRef.Tag,
-					PreferArch: preferArch,
-					PreferOS:   preferOS,
-					ImportMode: spec.ImportPolicy.ImportMode,
-					Image:      cache[tag],
-				})
+				if image, ok := cache[tag]; ok {
```
tkashem, line 234: what should we do if  cache is a miss?

tkashem: unit tests fail, if i do this:
```
@@ -230,13 +231,18 @@ func (imp *ImageStreamImporter) importImages(ctx context.Context, isi *imageapi.
 			tag.value = defaultRef.Tag
 			tags[tag] = append(tags[tag], i)
 			if len(tags[tag]) == 1 {
-				repo.Tags = append(repo.Tags, importTag{
+				image, ok := cache[tag]
+				tag := importTag{
 					Name:       defaultRef.Tag,
 					PreferArch: preferArch,
 					PreferOS:   preferOS,
 					ImportMode: spec.ImportPolicy.ImportMode,
-					Image:      cache[tag],
-				})
+					Image:      image,
+				}
+				if !ok {
+					tag.Err = errors.New("the referenced image in the import tag was not found in cache")
+				}
+				repo.Tags = append(repo.Tags, tag)
```

So i guess, I can't set `Err` for the tag.

flavianmissi: From what I can tell the cache miss here is expected. Down below on line 259 the cache will be populated.
I think the root cause of the panic is that when importing the image from upstream, `importManifest` seems to not be returning an error: https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/importer/importer.go?plain=1#L859
I'll see if I can recreate this on a unit test.

flavianmissi: @tkashem I think we just need to set the error down below, after the call to `importRepositoryFromDocker` (which is where the cache gets populated). This patch passes tests for me:

```
@@ -256,6 +257,14 @@ func (imp *ImageStreamImporter) importImages(ctx context.Context, isi *imageapi.
                                cache[j] = tag.Image
                        }
                        for _, index := range tags[j] {
+                               // we have seen this happen in our internal CI clusters. it's likely
+                               // caused by outages in the upstream registry, but we're not sure
+                               // exactly how we end up in this situation where an error isn't set.
+                               // to protect against that we throw up a generic error when it happens.
+                               // See https://issues.redhat.com/browse/OCPBUGS-35036 for details.
+                               if tag.Image == nil && tag.Err == nil {
+                                       tag.Err = errors.New("unknown error while importing image, please try again.")
+                               }
                                if tag.Err != nil {
                                        setImageImportStatus(isi, index, tag.Name, tag.Err)
                                        continue
```
Can you add this change to this PR? I believe we don't need this check here if we go with this suggestion.

flavianmissi:
I put together a different fix @tkashem : https://github.com/openshift/openshift-apiserver/pull/448
Let's close this one.
/close

PR #440, opened 2024-07-10
Add allow host users

haircommander:
Carrying a bunch, using this to test in a cluster


ibihim (commented):
/lgtm

benluddy:
I was under the impression that SCC is served by kube-apiserver via CRD on all flavors these days (https://github.com/openshift/kubernetes/blob/a3f31a4f2c587420f1ddb3bcdb161bf64a8c436b/openshift-kube-apiserver/enablement/intialization.go#L63). Do we have a reason to keep maintaining this?

haircommander:
yeah it seems we actually don't need this after all...

/close

PR #441, opened 2024-08-06
API-1820: Bump 1.30

dinhxuanvu:


go.mod
```diff
@@ -146,36 +148,35 @@ require (
...
 	golang.org/x/time v0.3.0 // indirect
-	golang.org/x/tools v0.16.1 // indirect
+	golang.org/x/tools v0.18.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20230803162519-f966b187b2e5 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20230726155614-23370e0ffb3e // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230822172742-b8732ec3820d // indirect
 	google.golang.org/grpc v1.58.3 // indirect
-	google.golang.org/protobuf v1.31.0 // indirect
+	google.golang.org/protobuf v1.33.0 // indirect
```
tkashem, line 169: this will fix: https://issues.redhat.com/browse/OCPBUGS-30528

CVE: https://pkg.go.dev/vuln/GO-2024-2611 (v1.33.0 fixes it)

p0lyn0mial:
`ci/prow/e2e-aws-ovn-builds` should be green before we merge this PR, we need to wait for https://github.com/openshift/origin/pull/29134

ardaguclu:
Failures look unrelated and the e2e-aws-ovn-builds should be fixed
/retest

hack/update-generated-conversions.sh
```diff
@@ -35,12 +39,20 @@ ALL_PEERS=(
...
     k8s.io/api/core/v1
     k8s.io/kubernetes/pkg/apis/core/v1
 )
 
+# Temporarily change the build tag so that the conversion funcs in the file can be loaded
+sed -i "s/$K8S_TAG/$TEMP_TAG/g" "$PATH_TO_FILE"
+
 echo "Generating conversions"
-${GOPATH}/bin/conversion-gen --input-dirs $(codegen::join , "${ALL_FQ_APIS[@]}") --extra-peer-dirs $(codegen::join , "${ALL_PEERS[@]}") --extra-dirs $(codegen::join , "${ALL_EXTRA_DIRS[@]}") --build-tag=ignore_autogenerated_openshift -O zz_generated.conversion --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt --v=8 "$@"
+${GOPATH}/bin/conversion-gen $(codegen::join_with_space , "${ALL_FQ_APIS[@]}") --extra-peer-dirs $(codegen::join , "${ALL_PEERS[@]}") --extra-dirs $(codegen::join , "${ALL_EXTRA_DIRS[@]}") --output-file zz_generated.conversion.go --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt --v=8 "$@" || true
```
p0lyn0mial, line 55: I think we would like to fail when the conversion binary fails. Please remove `|| true`

hack/update-generated-conversions.sh
```diff
@@ -35,12 +39,20 @@ ALL_PEERS=(
...
+    k8s.io/kubernetes/pkg/apis/core/v1
 )
 
 ALL_EXTRA_DIRS=(
     k8s.io/api/core/v1
     k8s.io/kubernetes/pkg/apis/core/v1
 )
 
+# Temporarily change the build tag so that the conversion funcs in the file can be loaded
+sed -i "s/$K8S_TAG/$TEMP_TAG/g" "$PATH_TO_FILE"
```
p0lyn0mial, line 52: I don't understand why we need to update the build tag on that file. Can you please explain?

hack/update-generated-conversions.sh
p0lyn0mial, line 55: why did we drop  `--build-tag=ignore_autogenerated_openshift` flag ?

hack/update-generated-conversions.sh
```diff
@@ -2,14 +2,18 @@
...
+
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../../../k8s.io/code-generator)}
 
-verify="${VERIFY:-}"
-
 GOFLAGS="" go install -mod=vendor ./${CODEGEN_PKG}/cmd/conversion-gen
 
 function codegen::join() { local IFS="$1"; shift; echo "$*"; }
+function codegen::join_with_space() { local IFS=" "; shift; echo "$*"; }
```
p0lyn0mial, line 16: why is this required ?

benluddy: It isn't, please replace `$(codegen::join_with_space , "${ALL_FQ_APIS[@]}")` with `"${ALL_FQ_APIS[@]}"` in the invocation at the bottom instead.

hack/update-generated-conversions.sh
```diff
@@ -35,12 +39,20 @@ ALL_PEERS=(
...
     k8s.io/apimachinery/pkg/types
     k8s.io/kubernetes/pkg/apis/core
+    k8s.io/api/core/v1
+    k8s.io/kubernetes/pkg/apis/core/v1
 )
 
 ALL_EXTRA_DIRS=(
     k8s.io/api/core/v1
     k8s.io/kubernetes/pkg/apis/core/v1
 )
```
benluddy, line 49: I see that `--extra-dirs` is no longer used by conversion-gen, even though the flag is wired. I don't fully understand the implications of moving these into ALL_PEERS. ALL_EXTRA_DIRS and the --extra-dirs flag in the invocation should be dropped from the script if they are pointless, though.

As an aside, it's surprising that conversion-gen doesn't complain about unrecognized flags. Is there a reason it doesn't?

hack/update-generated-conversions.sh
benluddy: I see that the option to flag the build tag for generated files is gone. I _think_ the purpose of the distinct build tag is to be able to find and use upstream-generated conversions without regenerating them in the process? I think I would prefer to see either a case made for reintroducing the ability to do this upstream, or even temporarily carrying a patch in our fork with an updated constant, over working around it in individual codegen scripts this way (keeping in mind that some similar version of these scripts is maintained across many repositories).

I thought we might have been able to override the string at link time with the `-X` flag, but it's a constant and not a variable, so the linker runs too late to be able to do anything about it.

hack/update-generated-conversions.sh
benluddy, line 55: Did the original reason for passing "$@" to this become obsolete? I don't see any arguments passed to this script from anywhere.

hack/verify-generated-conversions.sh
```diff
@@ -3,4 +3,14 @@
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
-VERIFY=--verify-only ${SCRIPT_ROOT}/hack/update-generated-conversions.sh
+${SCRIPT_ROOT}/hack/update-generated-conversions.sh
```
benluddy, line 6: How long have these verify scripts been broken?

hack/update-generated-conversions.sh
benluddy: https://github.com/openshift/openshift-apiserver/pull/441#discussion_r1790540950

hack/update-generated-conversions.sh
dinhxuanvu: ALL_PEERS and ALL_EXTRA_DIRS are ultimately merged together in the end so it is fine to combine them together. These flags you mentioned still exist so that's why it doesn't err out. The extra-dirs flag exists but it is just not wired.

hack/update-generated-conversions.sh
dinhxuanvu: All of tags that are removed in this PR is because those tags are deprecated on upstream especially since the conversion to gengo/v2 which no longer allows the custom build tag injection. The conversion-gen in particular replies on the conversion funcs from `k8s.io/kubernetes/pkg/apis/core/v1` in the conversion.go file. However, gengo will not load the conversion funcs from any conversion.go files that have `ignore_autogenerated` tag hence the temporary build tag modification.
I can see a case for reintroducing this feature back on upstream. However, given how unstaffed generator/gengo repos are, most of unimportant features are ignored and sit stale for years. Most of other projects who got impacted by this change are just coping those conversion funcs over to their repos. I find that solution too tedious and unmaintainable. This build tag modification solution is much better. We can use this until we sort it out upstream (if ever).

dinhxuanvu (commented):
Updated. PTAL

dinhxuanvu:
@p0lyn0mial @benluddy PTAL

hack/update-generated-conversions.sh
benluddy: > This build tag modification solution is much better.

Is that an approach we want to roll out to all of the similar codegen scripts in other openshift repositories (https://github.com/search?q=org%3Aopenshift+language%3AShell+%22conversion-gen%22+NOT+is%3Afork&type=code)?

> most of unimportant features are ignored

Is there an existing upstream issue for this one? It looks like rewiring a flag for this would only need to touch the conversion-gen subtree in k/k, not k/gengo. If you open a PR, I am sure it will be reviewed.

hack/update-generated-conversions.sh
benluddy: > The extra-dirs flag exists but it is just not wired.

That would be good to fix or at least report upstream.

dinhxuanvu: Agreed. I will ask if the intention is to remove that flag completely then I can open a PR for it.

hack/update-generated-conversions.sh
benluddy: Can we make the dependency on the generated corev1 conversions explicit by wrapping them with manual conversion funcs?

p0lyn0mial: How many functions would we need to copy?

I would at least create a PR upstream for discussion.

dinhxuanvu: I voice my strong opposition against doing the conversion func coping. It requires at least 48 funcs to be copied over based on the initial estimation along with some potential code changes to fix all of import alias inconsitencies. Those are a few thousands lines of codes that we have to carry over and maintain. This approach creates more code changes and introduce more code maintaince especially in case of code changes upstream because we still keep in sync due to k8s.io/kubernetes dependencies. Plus, if we introduct new fields in existing APIs, we potentially have to copy over more conversion funcs which lead to more maintainance efforts. In my opinion, doing all of this is entirely unnecessary and doesn't provide any advantages over my approach.

If we want to do a refactoring of the OAS code to get rid of unwanted dependencies, that is fine and I'm not opposing to that idea. However, that is a separate concern/work that is outside of the scope of this rebase.

My approach of temporary build-tag modification is minimal change and it simply restores the existing behavior that we already have pre-1.30. This solution will work for short/medium/long term without introducing any maintainance work. If we want to make an effort to not doing this solution, then that effort should be on upstream to reintroduce the build-tag which was removed as the maintainer deemed it was unused and untested. We can make a case for that. This approach is still appropreciate for short/meduim term why waiting for upstream discussion.

p0lyn0mial: I would prefer to restore the ability to pass custom build tags upstream, or carry a patch in our fork (o/k/conversion-gen) that would enable this functionality.

@deads2k, @sjenning  I would appreciate your opinion. I think it would help us make some progress here. 
It seems that upstream switched to gengo/v2 and modified conversion-gen in a way that no longer supports passing custom build tags. The purpose of a distinct build tag for OAS appears to have been to find and use conversions generated upstream. With this option no longer available, for example the generated corev1 conversions are missing after running the conversion script.

p0lyn0mial:
We agreed that before applying or maintaining custom patches, we would like to first bring back the removed build tag upstream. Restoring the build tag would allow us to drop the custom code we would otherwise have to maintain.

I had a chat with sttts last week, and he is okay with restoring the build tag. Changing the upstream binary doesn't require major changes.

dinhxuanvu:
@p0lyn0mial Updated the PR with a carry commit and reverse some changes in separate commits for now until we resolve the upstream issue. Will rebase later.

p0lyn0mial:
https://github.com/kubernetes/kubernetes/pull/128259 has been merged and will be available in 1.32+. 

To properly integrate the patch, we would need to move it to https://github.com/openshift/kubernetes-code-generator and then use this repository in OAS.

I don't thinks it's worth the effort, so I'm fine with using the temporary hack for the next two releases.
@dinhxuanvu , please restore your hack, add a comment explaining that it's a temporary solution, and indicate that it will be removed in 1.32.  Please also include instructions on what exactly needs to be removed.

dinhxuanvu:
@p0lyn0mial Done.

ardaguclu:
Unrelated;
/test e2e-aws-ovn-upgrade

hack/update-generated-openapi.sh
```diff
@@ -8,6 +8,7 @@ CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/kube-op
 GOFLAGS="" go install -mod=vendor ./${CODEGEN_PKG}/cmd/openapi-gen
 
 function codegen::join() { local IFS="$1"; shift; echo "$*"; }
+function codegen::join_with_space() { local IFS=" "; shift; echo "$*"; }
```
p0lyn0mial, line 11: does https://github.com/openshift/openshift-apiserver/pull/441#discussion_r1790519820 apply here as well?

dinhxuanvu: I need to join 2 separate lists together. So that's why this func is needed. Without it, it seems to err out on `invalid char ' '`.

benluddy: You can delete this function, delete L40-41 (the reassignments to KUBE_INPUT_DIRS and ORIGIN_INPUT_DIRS), and then replace L50 with: `"${KUBE_INPUT_DIRS[@]}" "${ORIGIN_INPUT_DIRS[@]}" \` and it should work.

benluddy: IIUC after the reassignments in L40-41, neither variable holds an array anymore and so the word-splitting behavior that we want from `"${KUBE_INPUT_DIRS[@]}"` is the same as if you had a single-element array with all of the package names joined by a space. (https://www.gnu.org/software/bash/manual/html_node/Arrays.html)

benluddy:
> I don't thinks it's worth the effort, so I'm fine with using the temporary hack for the next two releases.
@dinhxuanvu , please restore your hack, add a comment explaining that it's a temporary solution, and indicate that it will be removed in 1.32. Please also include instructions on what exactly needs to be removed.

:+1: That sounds fine. It's hard to imagine a patch that would be backported that would also break the workaround.

dinhxuanvu:
Updated and ready to go.

p0lyn0mial:
/lgtm

/hold
Since @benluddy also reviewed this PR, I'm holding it so he can have a final look.

hack/update-generated-conversions.sh
```diff
@@ -35,12 +38,19 @@ ALL_PEERS=(
...
+# Temporarily change the build tag so that the conversion funcs in the file can be loaded
+# This is a temporary solution due to the loss of --build-tag flag which is restored
+# in k8s 1.31 (see kubernetes/kubernetes#128259). When openshift-apiserver is
+# rebased to 1.32, the custom build tag `--build-tag=ignore_autogenerated_openshift`
+# will be added back to conversion-gen cmd.
+sed -i "s/$K8S_TAG/$TEMP_TAG/g" "$PATH_TO_FILE"
+
 echo "Generating conversions"
-${GOPATH}/bin/conversion-gen --input-dirs $(codegen::join , "${ALL_FQ_APIS[@]}") --extra-peer-dirs $(codegen::join , "${ALL_PEERS[@]}") --extra-dirs $(codegen::join , "${ALL_EXTRA_DIRS[@]}") --build-tag=ignore_autogenerated_openshift -O zz_generated.conversion --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt --v=8 "$@"
+${GOPATH}/bin/conversion-gen "${ALL_FQ_APIS[@]}" --extra-peer-dirs $(codegen::join , "${ALL_PEERS[@]}") --output-file zz_generated.conversion.go --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt --v=8 "$@" || { echo "conversion-gen failed" >&2 && true; }
```
p0lyn0mial, line 53: actually, just a small nit, I think we could get rid of `|| { echo "conversion-gen failed" >&2 && true; }` here and in the other places, right ?

PR #451, opened 2024-09-26
OCPBUGS-22332: Bump openshift/api for custom-host permission error message

Miciah:
Bump to github.com/openshift/library-go@354b673304be747827a7873426952da68f96b5c5, which improves the error message when the user tries to update a route's `spec.host`, `spec.subdomain`, or `spec.tls` field without having the required "custom-host" permission.  Instead of "field is immutable", the error message will have "you do not have permission".  This does not change the HTTP status code for these errors.

---

This vendors the change in https://github.com/openshift/library-go/pull/1677.


Miciah:
[The verify job failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_openshift-apiserver/451/pull-ci-openshift-openshift-apiserver-master-verify/1839293063717982208) because the `openapi-gen` command in the `update-generated-openapi.sh` script failed with the error message `unknown flag: --build-tag`.

lihongan:
I think we could close this one since it is superseded by https://github.com/openshift/openshift-apiserver/pull/459

PR #453, opened 2024-10-02
Bump 1.30

dinhxuanvu:


p0lyn0mial:
@dinhxuanvu I think that the `ci/prow/e2e-aws-ovn-builds` job is permanently failing https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-openshift-apiserver-master-e2e-aws-ovn-builds.

Open a new PR without any changes and trigger the job. If it fails then we should reach out to the owners (https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/OWNERS?) and let them now the job is not working.

dinhxuanvu:
No longer needed. See #441 instead.
/close

PR #474, opened 2025-01-09
OCPBUGS-47277: Bump golang x net 4.12

ingvagabund:
SSIA


ingvagabund:
- `golang.org/x/net/html` imported by `vendor/k8s.io/apimachinery/pkg/util/proxy/transport.go`
- html.NewTokenizer -> html.NewTokenizerFragment: none of the underlying methods/functions invoke the affected code
- html.ErrorToken (TokenType), html.StartTagToken (TokenType), html.SelfClosingTagToken (TokenType), TokenType = uint32, String() is the only method and does not invoke the affected code

PR #477, opened 2025-01-10
Prevent panic on missing tag image during import

deads2k:
This does not make the code work, it only makes the code avoid panicking when required data is missing.  Something else will likely break later.

### DIFF
--- a/pkg/image/apiserver/importer/importer.go
+++ b/pkg/image/apiserver/importer/importer.go
@@ -397,6 +397,11 @@ func (imp *ImageStreamImporter) importFromRepository(ctx context.Context, isi *i
 		}
 		status.Images[i].Status.Status = metav1.StatusSuccess
 
+		if tag.Image == nil {
+			status.Images[i].Status = invalidStatus("", field.Invalid(field.NewPath("tags").Key(tag.Name), nil, fmt.Sprintf("nil image for tag index=%d, tag name=%v", i, tag.Name)))
+			continue
+		}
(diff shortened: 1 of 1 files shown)


pkg/image/apiserver/importer/importer.go
```diff
@@ -397,6 +397,11 @@ func (imp *ImageStreamImporter) importFromRepository(ctx context.Context, isi *i
 		}
 		status.Images[i].Status.Status = metav1.StatusSuccess
 
+		if tag.Image == nil {
+			status.Images[i].Status = invalidStatus("", field.Invalid(field.NewPath("tags").Key(tag.Name), nil, fmt.Sprintf("nil image for tag index=%d, tag name=%v", i, tag.Name)))
```
Prashanth684, line 401: should the failures variable be increased to mark this as an import failure?

pkg/image/apiserver/importer/importer.go
```diff
@@ -397,6 +397,11 @@ func (imp *ImageStreamImporter) importFromRepository(ctx context.Context, isi *i
...
+			continue
+		}
+
```
ricardomaraschini, line 404: we may want to use the same messaging used for the same problem but in a [different code branch](https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/importer/importer.go#L260-L275).

deads2k:
example helped, got resolved.

PR #479, opened 2025-01-13
OCPBUGS-45861: add audit annotation to known issue

ricardomaraschini:
during image creation the apiserver logs:

```
unable to parse manifest for "sha256": unexpected end of JSON input
```

we have seen this issue happening during our ci and we want to increase its visibility. the idea is to do that using an audit annotation. with it we can then later filter these events and properly asses how often this is happening and, more importantly, when this issue is fixed.

### DIFF
--- a/pkg/image/apiserver/registry/image/strategy.go
+++ b/pkg/image/apiserver/registry/image/strategy.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/audit"
 	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/kubernetes/pkg/api/legacyscheme"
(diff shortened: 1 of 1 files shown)


pkg/image/apiserver/registry/image/strategy.go
```diff
@@ -19,8 +20,14 @@ import (
...
 
-// managedSignatureAnnotation used to be set by image signature import controller as a signature annotation.
-const managedSignatureAnnotation = "image.openshift.io/managed-signature"
+const (
+	// managedSignatureAnnotation used to be set by image signature import
+	// controller as a signature annotation.
+	managedSignatureAnnotation = "image.openshift.io/managed-signature"
+	// reasonAnnotation is a audit annotation we add to the context
+	// when we see a failure during an image processing.
+	reasonAnnotation = "image.openshift.io/reason"
```
deads2k, line 29: `image.internal.openshift.io/reason` please.  We'll change this without regard for backward compatbility if we find we need to.

ricardomaraschini: fixed.

flavianmissi:
The aws e2e failures are due to timeout waiting for cluster install, which seems unlikely to be related to the changes in this PR.
/retest

ricardomaraschini:
I will be closing this for now. I still plan to reopen once it but this has a lower priority when compared with solving the actual bug.

PR #480, opened 2025-01-14
OCPBUGS-45861: move on to the next digest/tag during failures

ricardomaraschini:
if we fail to process a given tag or digest we should move to the next and not return. if we "prematurely" return we may cause some of the tags or digest have their Image pointer nulled, possibly causing a panic.

### DIFF
--- a/pkg/image/apiserver/importer/importer.go
+++ b/pkg/image/apiserver/importer/importer.go
@@ -791,7 +791,7 @@ func (imp *ImageStreamImporter) importRepositoryFromDocker(ctx context.Context, 
 					"unable to import manifest list %q: %v",
 					ref.Exact(), err)
 				importDigest.Err = err
-				return
+				continue
 			}
 			importDigest.Manifests = images
 		}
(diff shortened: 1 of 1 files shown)


pkg/image/apiserver/importer/importer.go
```diff
@@ -791,7 +791,7 @@ func (imp *ImageStreamImporter) importRepositoryFromDocker(ctx context.Context,
 					"unable to import manifest list %q: %v",
 					ref.Exact(), err)
 				importDigest.Err = err
-				return
+				continue
```
Prashanth684, line 794: even if we continue, the image for this digest (and similarly for the tag below) is going to be null , correct? wouldn't that still be an issue?

ricardomaraschini:
@flavianmissi IIUC we also need the approval from someone in the OWNERS file. The [message](https://github.com/openshift/openshift-apiserver/pull/480#issuecomment-2593366200) is kind of dubious so I am not sure it actually needs to be Ben Parees. 

/assign @flavianmissi

flavianmissi:
`e2e-aws-ovn-serial` seems specially stubborn - the failures are unrelated to the changes in this PR so I'm going to override it.

/override ci/prow/e2e-aws-ovn-serial

neisw:
/override ci/prow/e2e-aws-ovn-serial

Failure is being investigated and present outside this pr.

PR #483, opened 2025-01-21
OCPBUGS-45861: validate image property isn't nil before using

ricardomaraschini:
even though we do not expect these two properties to be nil (Image and Err) we better check them to avoid having panics if something changes in the future.

this pr introduces the same guardrails used in two different parts of the code: makes sure we have at least one of `Err` or `Image` properties set.


ricardomaraschini:
Judging by the [history](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-openshift-apiserver-master-okd-scos-e2e-aws-ovn) the `ci/prow/okd-scos-e2e-aws-ovn` job seems to be a little bit flaky. As I have seen some successes on the history let's give this one another shot.

/test okd-scos-e2e-aws-ovn

PR #464, opened 2024-12-13
OCPBUGS-45729: Updating ose-openshift-apiserver-container image to be…

gangwgr:
… consistent with ART for 4.19

PR #486, opened 2025-01-27
[WIP]Updating ose-openshift-apiserver-container image to be consistent with ART for 4.19

wangke19:


gangwgr:
@dgrisonnet We tried that https://github.com/openshift/openshift-apiserver/pull/464, in which pr test are not passing

dgrisonnet:
Looking at the failure on the other PR, you need to run `hack/update-generated-conversions.sh` and push the changes to fix the test

gangwgr:
We tried that, but without library update that was failing

wangke19:
> Looking at the failure on the other PR, you need to run `hack/update-generated-conversions.sh` and push the changes to fix the test

Done, let's see restuls of ci jobs.

dgrisonnet:
Seems like they are still some failures, I'll try to take some time to investigate

wangke19:
Depends on https://github.com/openshift/kubernetes-apiserver/pull/65

dinhxuanvu:
@wangke19 You will need to update the go.mod to reference my commit in this PR. This PR will serve as a proof for k8s-apiserver PR. If this PR passes CI, then k8s-apiserver PR can be merged.

wangke19:
@dinhxuanvu I updated k8s.io/apiserver reference of openshift/kubernetes-apiserver#65, will run tests to look the results.

wangke19:
@dinhxuanvu Do you have any suggestions for failed unit tests?

pkg/apps/apis/apps/validation/validation.go
```diff
@@ -258,7 +260,34 @@ func validateDeploymentStrategy(strategy *appsapi.DeploymentStrategy, pod *kapi.
...
+	if strategy.Resources.Limits != nil {
+		for resourceName, quantity := range strategy.Resources.Limits {
+			quota := &kapi.ResourceQuota{
+				Spec: kapi.ResourceQuotaSpec{
+					Hard: map[kapi.ResourceName]resource.Quantity{
+						resourceName: quantity,
+					},
+				},
+			}
+			if errs = kapivalidation.ValidateResourceQuota(quota); len(errs) > 0 {
```
dinhxuanvu, line 272: I would like to ask why you added this block of code to validate resource limits? The func `ValidateResourceQuota` actually validate ObjectMeta as well which leads to all of failures on unit tests due to the fact this RQ object doesn't have name or namespace values which are required.

Also please separate the vendor code change to one commit and your own changes to another commit. This commit is so big. It is very hard to review as browser can't load such a big diff.

dinhxuanvu (changes requested):
One of your commits causes failures on unit tests unfortunately. See the comment for details.

pkg/apps/apis/apps/validation/validation.go
wangke19: To fix build errros, I added some funcs. I have removed them now.

wangke19:
@dinhxuanvu looks good, Wait for https://github.com/openshift/kubernetes-apiserver/pull/65 merge.

dinhxuanvu:
@dgrisonnet Would you mind taking a look and blessing it with an approval? Thanks.

pkg/apps/apis/apps/validation/validation.go
```diff
@@ -258,8 +258,6 @@ func validateDeploymentStrategy(strategy *appsapi.DeploymentStrategy, pod *kapi.
 		}
 	}
 
-	errs = append(errs, validation.ValidateResourceRequirements(&strategy.Resources, podClaimNames, fldPath.Child("resources"), kapivalidation.PodValidationOptions{})...)
```
dgrisonnet, line 261: I don't understand this change, why do we need it?

wangke19: This func `ValidateResourceRequirements`  doesn't exist in https://pkg.go.dev/k8s.io/kubernetes@v1.32.1/pkg/apis/core/validation.

wangke19: Updated.

dgrisonnet (approved):
/lgtm
/approve

dgrisonnet (approved):

go.mod
```diff
@@ -151,72 +151,73 @@ require (
...
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
 replace (
 	github.com/distribution/distribution/v3 => github.com/openshift/docker-distribution/v3 v3.0.0-20240215131201-6b2f5d2f1f43
 	github.com/docker/docker => github.com/openshift/moby-moby v0.0.0-20190308215630-da810a85109d
-	k8s.io/api => k8s.io/api v0.31.1
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.1
-	k8s.io/apimachinery => k8s.io/apimachinery v0.31.1
+	k8s.io/api => k8s.io/api v0.32.1
```
Miciah, line 195: It is confusing to me that you bump to `k8s.io/api v0.32.3` above but keep the `k8s.io/api => k8s.io/api v0.32.1` rule here.  Is this `replace` rule still needed?

Miciah: I'm not aware of a problem with 0.32.1, but the `replace` rule could cause confusion about what which version we're actually using and could trip up the next person to do a version bump.

PR #508, opened 2025-04-30
OCPBUGS-38120: project: run validation function when deleting an object

vrutkovs:
Ensure that validation admission hooks are being run when project is being deleted

### DIFF
--- a/pkg/project/apiserver/registry/project/proxy/proxy.go
+++ b/pkg/project/apiserver/registry/project/proxy/proxy.go
@@ -214,5 +214,26 @@ func (s *REST) Delete(ctx context.Context, name string, objectFunc rest.Validate
 	if options != nil {
 		opts = *options
 	}
+	if objectFunc != nil {
+		obj, err := s.Get(ctx, name, &metav1.GetOptions{})
+		if err != nil {
+			return nil, false, err
+		}
+		projectObj, ok := obj.(*projectapi.Project)
+		if !ok || projectObj == nil {
+			return nil, false, fmt.Errorf("not a project: %#v", obj)
+		}
+
+		// Make sure the object hasn't changed between Get and Delete - pass UID and RV to delete options
+		if opts.Preconditions == nil {
+			opts.Preconditions = &metav1.Preconditions{}
+		}
+		opts.Preconditions.UID = &projectObj.UID
+		opts.Preconditions.ResourceVersion = &projectObj.ResourceVersion
+
+		if err := objectFunc(ctx, obj); err != nil {
+			return nil, false, err
+		}
+	}
 	return &metav1.Status{Status: metav1.StatusSuccess}, false, s.client.Delete(ctx, name, opts)
 }
--- a/pkg/project/apiserver/registry/project/proxy/proxy_test.go
+++ b/pkg/project/apiserver/registry/project/proxy/proxy_test.go
@@ -1,6 +1,7 @@
 package proxy
 
 import (
+	"context"
 	"strings"
 	"testing"
 
@@ -8,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/authentication/user"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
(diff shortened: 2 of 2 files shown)


pkg/project/apiserver/registry/project/proxy/proxy.go
```diff
@@ -214,5 +214,12 @@ func (s *REST) Delete(ctx context.Context, name string, objectFunc rest.Validate
 	if options != nil {
 		opts = *options
 	}
+	obj, err := s.Get(ctx, name, &metav1.GetOptions{})
+	if err != nil {
+		return nil, false, err
+	}
+	if err := objectFunc(ctx, obj); err != nil {
```
benluddy, line 221: How will we guarantee that namespace that is deleted is the same namespace that the admitted project was generated from?

The proxied namespace delete request should have a UID precondition (to block deletion if the underlying namespace was deleted and re-created after it was observed here) and RV precondition (to block deletion if the underlying namespace was modified such that the generated project would no longer pass admission).

Do there need to be retries here?

vrutkovs: >to block deletion if the underlying namespace was deleted and re-created after it was observed here

I don't think its possible to delete and recreate a namespace without project being deleted too

>to block deletion if the underlying namespace was modified such that the generated project would no longer pass admission

The admission is applied to project details, not namespace's. I don't think namespace details matter here?

benluddy: The content of the project being admitted is entirely a function of the namespace with the same name -- it's a facade with no storage of its own.

For namespace deletes, the validation function is passed down to the storage layer, which reads the latest revision of the object, calls the validation function, and issues an optimistic delete that fails if the stored revision is not the same as the validated revision. If the revision has changed, it automatically retries in a loop.

Imagine you have an admission plugin that rejects deletion of projects with the label "nodelete", and you have a namespace named "foo" with no labels. You might have this sequence:

1. A "delete project 'foo'" request arrives at openshift-apiserver.
2. openshift-apiserver gets namespace "foo". It does not have any labels, so the project "foo" also does not have any labels, and it passes admission.
3. Request B "update namespace 'foo'" arrives and the "nodelete' label is added (changing both the namespace and the project named "foo").
4. openshift-apiserver sends its "delete namespace 'foo'" request to kube-apiserver and the namespace is deleted. The client that issued the request in step 1 sees 200 and the project "foo" and namespace "foo" are gone.

The "delete project 'foo'" and "update namespace 'foo'" requests can't be serialized in an order that makes sense. If the update request happened first, then admission should have rejected the delete request. If the delete request happened first, then the update request should have failed.

vrutkovs: gotcha, since we do Get now we can store RV/UID - and it needs to be passed to Delete option. Same needs to happen for Update (new issue required for this). Other resources are probably affected?

vrutkovs: Done

>Same needs to happen for Update (new issue required for this)

`meta.UpdateOptions` don't have UID/RV, so we can do an additional UID/RV check between Get and Update, but it probably won't be sufficient

benluddy: It's enough to set UID and ResourceVersion on the ObjectMeta of that object you pass to the update request.

pkg/image/apiserver/importer/import_integration_test.go
```diff
@@ -94,10 +94,11 @@ func TestImageStreamImportQuayIO(t *testing.T) {
 	importCtx := importer.NewStaticCredentialsContext(rt, nil, nil)
 
 	repositoryName := quayRegistryName + "/coreos/etcd"
+	tag := "v3.4.37"
 	imports := &imageapi.ImageStreamImport{
 		Spec: imageapi.ImageStreamImportSpec{
 			Images: []imageapi.ImageImportSpec{
-				{From: kapi.ObjectReference{Kind: "DockerImage", Name: repositoryName}},
+				{From: kapi.ObjectReference{Kind: "DockerImage", Name: fmt.Sprintf("%s:%s", repositoryName, tag)}},
```
everettraven, line 101: Was there a specific reason this needed changing?

everettraven (approved):
Other than a question about why the imagestream import needed modification, this looks good to me.

As long as tests are passing fine, I don't see a reason to block on my question.

pkg/project/apiserver/registry/project/proxy/proxy.go
```diff
@@ -214,5 +214,25 @@ func (s *REST) Delete(ctx context.Context, name string, objectFunc rest.Validate
 	if options != nil {
 		opts = *options
 	}
+	obj, err := s.Get(ctx, name, &metav1.GetOptions{})
```
benluddy, line 217: There's no need to make a roundtrip to kube-apiserver here when there is no project delete admission handler (and that seems to be the common case in practice).

pkg/image/apiserver/importer/import_integration_test.go
benluddy: This fails on master because it's not a hermetic test and `quay.io/coreos/etcd:latest` has ceased to exist.

@vrutkovs Could you pull the test fix into a standalone test fix PR?

vrutkovs: Created https://github.com/openshift/openshift-apiserver/pull/509

pkg/project/apiserver/registry/project/proxy/proxy.go
```diff
@@ -214,5 +214,26 @@ func (s *REST) Delete(ctx context.Context, name string, objectFunc rest.Validate
 	if options != nil {
 		opts = *options
 	}
+	if objectFunc != nil {
+		obj, err := s.Get(ctx, name, &metav1.GetOptions{})
+		if err != nil {
+			return nil, false, err
+		}
+		projectObj, ok := obj.(*projectapi.Project)
```
wangke19, line 222: This seems to be capturing the problem that occurs here https://github.com/openshift/origin/pull/29728

benluddy: The linked issue is a separate problem on the client side. Here, Get always returns the same type, but a type assertion is required because the return type in the method signature is the interface `runtime.Object`. The type assertion should never fail unless the implementation of the Get method breaks. The underlying request made by Get is actually to the "namespaces" resource.

benluddy:
/lgtm

Would be great to see an integration test similar to https://github.com/kubernetes/kubernetes/blob/b20fca7484f0a960d46400d936e01a2728a3672e/test/integration/apiserver/admissionwebhook/admission_test.go#L454 in openshift-apiserver to make sure it runs admission for all the requests we expect. Can we track that somewhere please?

vrutkovs:
Created https://issues.redhat.com/browse/CNTRLPLANE-930

PR #509, opened 2025-05-20
OCPBUGS-54458: Fix image reference in TestImageStreamImportQuayIO

vrutkovs:
quay.io/coreos/etcd:latest is gone, use a different tag instead


wangke19:
Before, we have a pr for this https://github.com/openshift/openshift-apiserver/pull/503

vrutkovs:
@wangke19 thanks, I closed the duplicate bug and used the previous one instead.

I don't think its a good idea to use coreos-operator - and tagged version of etcd is less likely to be removed imo

wangke19:
Agree with you, I did the same in a commit https://github.com/openshift/openshift-apiserver/pull/486/commits/8ddda1abf307714f1e8a589a4955455a54d521a0.
lgtm

QiWang19:
Can we backport this bug fix to 4.17, since the test failure blocks the backport PR https://github.com/openshift/openshift-apiserver/pull/500

PR #527, opened 2025-06-17
OCPBUGS-57474: Requests inappropriately skipped + test

jacobsee:
Previously, a request that removed a role or rolebinding from a subject for which we had a cached result would be skipped unless a clusterrole/clusterrolebinding was also modified which would force a full cache invalidation. This adds more conditions to the skip checker.

cc: @benluddy

### DIFF
--- a/pkg/project/auth/cache.go
+++ b/pkg/project/auth/cache.go
@@ -545,6 +545,13 @@ func skipReview(request *reviewRequest, lastKnownValue *reviewRecord) bool {
 		return false
 	}
 
+	// if the request has no role binding or role information but we previously had subjects, force review to detect removals
+	if len(request.roleBindingUIDToResourceVersion) == 0 && len(request.roleUIDToResourceVersion) == 0 {
+		if len(lastKnownValue.users) > 0 || len(lastKnownValue.groups) > 0 {
+			return false
+		}
+	}
+
 	// if you see a new role binding, or a newer version, we need to do a review
 	for k, v := range request.roleBindingUIDToResourceVersion {
 		oldValue, exists := lastKnownValue.roleBindingUIDToResourceVersion[k]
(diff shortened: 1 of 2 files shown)


pkg/project/auth/cache.go
```diff
@@ -545,6 +545,13 @@ func skipReview(request *reviewRequest, lastKnownValue *reviewRecord) bool {
 		return false
 	}
 
+	// if the request has no role binding or role information but we previously had subjects, force review to detect removals
+	if len(request.roleBindingUIDToResourceVersion) == 0 && len(request.roleUIDToResourceVersion) == 0 {
+		if len(lastKnownValue.users) > 0 || len(lastKnownValue.groups) > 0 {
+			return false
+		}
+	}
+
```
benluddy, line 554: Doesn't this match every namespace review request in practice, assuming we have at least one subject that is authorized for all namespaces (`synchronizeNamespaces` appears to always submit a request with nil role and rolebinding maps)? I can delete the rest of the patch and the test passes anyway.

pkg/project/auth/cache.go
```diff
@@ -560,6 +567,21 @@ func skipReview(request *reviewRequest, lastKnownValue *reviewRecord) bool {
 			return false
 		}
 	}
+
+	// if you see a role binding that was previously known but is no longer in the request, we need to do a review
+	for oldUID := range lastKnownValue.roleBindingUIDToResourceVersion {
+		if _, exists := request.roleBindingUIDToResourceVersion[oldUID]; !exists {
+			return false
+		}
+	}
```
benluddy, line 576: IIUC, on every sync, there's one request per role or rolebinding, and each request will contain exactly one entry in the appropriate UID:RV map, and the "last known" value contains the result of merging all of the UID:RV entries and namespace RVs it has seen.

Won't this guarantee 1 review per role and 1 review per rolebinding assuming we have more than one of each in every namespace?

jacobsee:
Closing this out in favor of a rethinking of the cache

PR #528, opened 2025-06-19
WRKLDS-1599: Drop support for image manifest schema 1

tchap:
The change itself is fairly small, but there were quite a few tests using schema1. These had to be updated. Some test cases also don't make sense any more as you cannot really update an image schema2, that would change the hash. So tests updating an image with a signature were removed as signatures are not supported in schema2 implicitly.

Feel free to propose more test cases if anything comes up your mind, I did what I could without learning all the intricacies of the codebase, which meant extending the tests for `imageutil.InternalImageWithMetadata` for all schema types, then aligning all other breaking tests.

The PR is a bit chaotic in the sense that there are multiple mock manifests and configs scattered across the codebase. I normalized what made sense to me, but some chaos is still there. Particularly there are separate tests for filling in the image object from the string manifest and config, then there are API tests for actually posting a manifest. These are entirely separate tests.


pkg/image/apiserver/internal/testutil/image.go
```diff
@@ -0,0 +1,69 @@
+package testutil
```
tchap, line 1: This is actually unchanged, just moved.

pkg/image/apiserver/registry/image/etcd/etcd_test.go
```diff
@@ -146,300 +137,169 @@ func TestWatch(t *testing.T) {
...
-				DockerImageManifest:  etcdManifest,
-			},
-		}}
+	}
 
 	for i, test := range testCases {
-		storage, server := newStorage(t)
-		defer server.Terminate(t)
-		defer storage.Store.DestroyFunc()
+		t.Run(test.name, func(t *testing.T) {
```
tchap, line 267: Added `t.Run` here and there, which makes the diff ugly, but helps for debugging.

tchap:
> These types are duplicated in [openshift/library-go pkg/image/dockerv1client/types.go](https://github.com/openshift/library-go/blob/master/pkg/image/dockerv1client/types.go)

I guess I can just migrate to library-go in a subsequent PR and remove the packages as present in this repo.

PR #530, opened 2025-06-24
WIP: OCPBUGS-57474: Authorization Cache V2

jacobsee:
- Move existing (periodic-sync based) `AuthorizationCache` from `cache.go` to `AuthorizationCacheV1` in `cachev1.go` (no longer instantiated anywhere currently)
- Add generic interface for an `AuthorizationCache` to `cache.go`
- Add `AuthorizationCacheV2`, which uses periodic cluster and namespace-specific hashing to track changes and invalidate the cache/notify watchers
- Add `ReactiveAuthorizationCacheV2`, which wraps `AuthorizationCacheV2` in event-handling so that heavy sync checks can happen less frequently and changes are still processed in a timely manner
- Switch the Project API to instantiating a `ReactiveAuthorizationCacheV2`
- Significant testing

Motivated by the fact that the current auth cache does not appear to be notifying properly when permissions have been removed during incremental synchronizations (https://issues.redhat.com/browse/OCPBUGS-57474), and it looks to be an old issue in a system that is rather difficult to reason around.

### DIFF
--- a/pkg/project/auth/cachev2_global_test.go
+++ b/pkg/project/auth/cachev2_global_test.go
@@ -124,24 +124,24 @@ func TestGlobalRBACCache_hasResourceVersionChanged(t *testing.T) {
 
 func TestGlobalRBACCache_updateResourceVersions(t *testing.T) {
 	tests := []struct {
-		name                       string
-		clusterRoleVersion         string
-		clusterRoleBindingVersion  string
+		name                      string
+		clusterRoleVersion        string
+		clusterRoleBindingVersion string
 	}{
 		{
-			name:                       "update both versions",
-			clusterRoleVersion:         "123",
-			clusterRoleBindingVersion:  "456",
+			name:                      "update both versions",
+			clusterRoleVersion:        "123",
+			clusterRoleBindingVersion: "456",
 		},
 		{
-			name:                       "empty cluster role version",
-			clusterRoleVersion:         "",
-			clusterRoleBindingVersion:  "456",
+			name:                      "empty cluster role version",
+			clusterRoleVersion:        "",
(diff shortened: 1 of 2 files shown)


pkg/project/auth/cachev2_global.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+// GetClusterRoleLister returns the cluster role lister
+func (g *GlobalRBACCache) GetClusterRoleLister() SyncedClusterRoleLister {
+	return g.clusterRoleLister
+}
+
+// GetCurrentHash returns the current global RBAC hash
+func (g *GlobalRBACCache) GetCurrentHash() string {
```
benluddy, line 43: Do callers need to be responsible for detecting that this hash might be stale and for updating it? I think correctness depends on always getting the latest hash. Caching based on the last sync RV makes sense as an optimization, but I think it's an implementation detail and not something callers care about.

pkg/project/auth/cachev1_test.go
```diff
@@ -129,7 +54,7 @@ func TestSyncNamespace(t *testing.T) {
 	nsIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	nsLister := corev1listers.NewNamespaceLister(nsIndexer)
 
-	authorizationCache := NewAuthorizationCache(
```
benluddy, line 132: Are there tests that can be run against both implementations? Or a shim that runs both implementations for real and detects behavior differences?

benluddy (commented):
Thanks for all your work on this, it looks really promising!

PR #532, opened 2025-07-09
WIP: WRKLDS-1599: Update to distribution/v3@v3.0.0

tchap:
Move to use upstream distribution/v3 instead of our private fork.

This is blocked by https://github.com/openshift/library-go/pull/1972


pkg/image/apiserver/importer/importer_imageindex_test.go
```diff
@@ -0,0 +1,36 @@
...
+	_ "embed"
+	"testing"
+
+	"github.com/distribution/distribution/v3/manifest/ocischema"
+)
+
+// imageIndexJSON actually contains the same manifests as manifestListJSON,
+// so the same manifests can be used during testing.
+//
+//go:embed testdata/image-index.json
```
flavianmissi: why did we need to create a copy of `manifestListJSON` ? 🤔

PR #543, opened 2025-08-26
OCPBUGS-57474: review rbac upon role/rolebinding removal

ricardomaraschini:
if a role or rolebinding is removed from a namespace we need to reassess the namespace (review its rbac).

instead of passing just one object (role or rolebinding) we are now passing them all so we can compare with the previously known state.

## Summary by CodeRabbit

* **Performance**
  * Improved policy synchronization by batching updates per namespace, reducing redundant processing and API calls.

* **Bug Fixes**
  * Ensures permission reviews rerun when roles or bindings change or are removed, preventing stale authorization results.
  * Preserves prior cache data only when new requests don’t provide it, avoiding unintended data loss.

* **Reliability**
  * More consistent and up-to-date authorization outcomes across namespaces.

* **Chores**
  * Internal synchronization logic refined with no changes to public APIs.

### DIFF
--- a/pkg/project/auth/cache.go
+++ b/pkg/project/auth/cache.go
@@ -313,11 +313,20 @@ func (ac *AuthorizationCache) synchronizePolicies(userSubjectRecordStore cache.S
 		utilruntime.HandleError(err)
 		return
 	}
+
+	reviewRequests := map[string]*reviewRequest{}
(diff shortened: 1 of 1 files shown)


ricardomaraschini:
As agreed with @benluddy I will be reopening this using a lower risk change. Closing this one.

PR #544, opened 2025-08-27
OCPBUGS-57474: ensure cache invalidation after a time

ricardomaraschini:
this commit ensures that we invalidate the cache at least every 15 seconds. this has been the chosen workaround for an existing bug that is on our code since long ago but just recently bubbled up in our e2e tests.

ideally we would refactor the whole cache layer but giving the time constraints we have opted-in for a more conservative approach.

## Summary by CodeRabbit

- Bug Fixes
  - Added a time-based TTL for the authorization cache so permissions refresh automatically (reducing stale access) while preserving existing detection of resource changes to ensure up-to-date role/permission behavior.
- Tests
  - Added unit tests validating TTL-based expiry behavior, that cache refresh occurs after lifespan expiry, and that refresh still triggers on resource changes.

### DIFF
--- a/pkg/project/auth/cache.go
+++ b/pkg/project/auth/cache.go
@@ -26,6 +26,11 @@ import (
 	authorizationapi "github.com/openshift/openshift-apiserver/pkg/authorization/apis/authorization"
 )
 
+// defaultMaxCacheLifespan is used to control how long the cache can live
+// without being invalidated. This is a workaround for a bug, see
+// https://issues.redhat.com/browse/OCPBUGS-57474 for further details.
+const defaultMaxCacheLifespan = 15 * time.Second
+
 // Lister enforces ability to enumerate a resource based on role
 type Lister interface {
 	// List returns the list of Namespace items that the user can access
@@ -197,6 +202,16 @@ type AuthorizationCache struct {
 
 	watchers    []CacheWatcher
 	watcherLock sync.Mutex
+
+	// lastCacheInvalidation and maxCacheLifespan are used together.
+	// We use them to control the maximum time allowed between cache
+	// invalidations. This is a workaround for a bug. We deemed the
+	// risk of refactoring the cache to be too high so close to the
+	// release date, so close to the release so here we are. For
+	// further details see:
+	// https://issues.redhat.com/browse/OCPBUGS-57474
+	lastCacheInvalidation time.Time
+	maxCacheLifespan      time.Duration
 }
 
 // NewAuthorizationCache creates a new AuthorizationCache
@@ -360,8 +375,32 @@ func (ac *AuthorizationCache) purgeDeletedNamespaces(oldNamespaces, newNamespace
 }
 
 // invalidateCache returns true if there was a change in the cluster namespace that holds cluster role and role bindings
-func (ac *AuthorizationCache) invalidateCache() bool {
-	invalidateCache := false
+func (ac *AuthorizationCache) invalidateCache() (invalidateCache bool) {
+	// XXX we ensure that we always invalidate the cache from time to time.
+	// Workaround for https://issues.redhat.com/browse/OCPBUGS-57474.
+	defer func() {
+		// we need to know how long is the cache interval. if none is
+		// provided then we must assume a default one.
+		if ac.maxCacheLifespan == 0 {
+			ac.maxCacheLifespan = defaultMaxCacheLifespan
+		}
+
+		// if either we are returning true or this is the first time we
+		// run the invalidateCache we need to set the cache interval
+		// start to now. we will start counting from now.
+		if invalidateCache || ac.lastCacheInvalidation.IsZero() {
+			ac.lastCacheInvalidation = time.Now()
+			return
+		}
+
+		// we only reach here if we are returning false. on this case
+		// we want to change the return to true if the interval has
+		// been exceeded.
+		if time.Since(ac.lastCacheInvalidation) > ac.maxCacheLifespan {
+			ac.lastCacheInvalidation = time.Now()
+			invalidateCache = true
+		}
+	}()
 
 	clusterRoleList, err := ac.clusterRoleLister.List(labels.Everything())
 	if err != nil {
@@ -397,9 +436,11 @@ func (ac *AuthorizationCache) invalidateCache() bool {
 
 // synchronize runs a a full synchronization over the cache data.  it must be run in a single-writer model, it's not thread-safe by design.
 func (ac *AuthorizationCache) synchronize() {
-	// if none of our internal reflectors changed, then we can skip reviewing the cache
+	// if none of our internal reflectors changed, and the cache hasn't
+	// expired yet we can skip reviewing the cache.
+	expired := time.Since(ac.lastCacheInvalidation) > ac.maxCacheLifespan
 	skip, currentState := ac.skip.SkipSynchronize(ac.lastState, ac.lastSyncResourceVersioner, ac.roleLastSyncResourceVersioner)
-	if skip {
+	if skip && !expired {
 		return
 	}
 
--- a/pkg/project/auth/cache.go
+++ b/pkg/project/auth/cache.go
@@ -203,13 +203,10 @@ type AuthorizationCache struct {
 	watchers    []CacheWatcher
 	watcherLock sync.Mutex
 
-	// lastCacheInvalidation and maxCacheLifespan are used together.
-	// We use them to control the maximum time allowed between cache
-	// invalidations. This is a workaround for a bug. We deemed the
-	// risk of refactoring the cache to be too high so close to the
-	// release date, so close to the release so here we are. For
-	// further details see:
-	// https://issues.redhat.com/browse/OCPBUGS-57474
+	// lastCacheInvalidation and maxCacheLifespan control the maximum time
+	// between cache invalidations. This is a temporary workaround due to
+	// release timing risk.
+	// See https://issues.redhat.com/browse/OCPBUGS-57474 for details.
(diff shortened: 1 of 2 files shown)


pkg/project/auth/cache_test.go
```diff
@@ -279,3 +280,93 @@ func TestInvalidateCache(t *testing.T) {
...
+		t.Errorf("expected false on check first check, got true")
+	}
+
+	// the last invalidation time should be unchanged.
+	if !ac.lastCacheInvalidation.Equal(start) {
+		t.Errorf("expected last invalidation time to be unchanged")
+	}
+
+	// give the lifespan time to expire.
+	time.Sleep(time.Second + 50*time.Millisecond)
```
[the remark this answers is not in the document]
ricardomaraschini: What is the reasoning behind this ? Is it because the test will take extra second ? I would like to understand.

pkg/project/auth/cache.go
```diff
@@ -360,8 +375,32 @@ func (ac *AuthorizationCache) purgeDeletedNamespaces(oldNamespaces, newNamespace
...
-	invalidateCache := false
+func (ac *AuthorizationCache) invalidateCache() (invalidateCache bool) {
+	// XXX we ensure that we always invalidate the cache from time to time.
+	// Workaround for https://issues.redhat.com/browse/OCPBUGS-57474.
+	defer func() {
+		// we need to know how long is the cache interval. if none is
+		// provided then we must assume a default one.
+		if ac.maxCacheLifespan == 0 {
+			ac.maxCacheLifespan = defaultMaxCacheLifespan
+		}
```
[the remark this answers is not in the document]
ricardomaraschini: That was my initial approach but then tests aren't using the `NewAuthorizationCache()` function and to make them use it will require more code changes. Before I dig into it, are you OK with making the tests call `NewAuthorizationCache()` ?  Do you have a different idea ?

pkg/project/auth/cache.go
```diff
@@ -360,8 +375,32 @@ func (ac *AuthorizationCache) purgeDeletedNamespaces(oldNamespaces, newNamespace
...
+
+		// if either we are returning true or this is the first time we
+		// run the invalidateCache we need to set the cache interval
+		// start to now. we will start counting from now.
+		if invalidateCache || ac.lastCacheInvalidation.IsZero() {
+			ac.lastCacheInvalidation = time.Now()
```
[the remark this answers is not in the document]
ricardomaraschini: Do you mean something like `invalidateCache(lastInvalidation time.Time)` ?

pkg/project/auth/cache_test.go
```diff
@@ -279,3 +280,93 @@ func TestInvalidateCache(t *testing.T) {
...
+	// invalidated.
+	if invalidate := ac.invalidateCache(); !invalidate {
+		t.Errorf("expected true after cache lifespan exceeded, got false")
+	}
+
+	// we check that the last invalidation time has been updated now that
+	// the max lifespan has been exceeded.
+	if ac.lastCacheInvalidation.Equal(start) {
+		t.Errorf("expected last invalidation time to be updated")
+	}
```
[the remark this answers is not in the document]
ricardomaraschini: OK, this can be easily addressed.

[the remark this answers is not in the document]
ricardomaraschini: Addressed, let me know if it is OK now.

pkg/project/auth/cache_test.go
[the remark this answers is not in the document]
ricardomaraschini: Addressed, let me know if it is OK now.

pkg/project/auth/cache.go
```diff
@@ -360,9 +375,12 @@ func (ac *AuthorizationCache) purgeDeletedNamespaces(oldNamespaces, newNamespace
 }
 
 // invalidateCache returns true if there was a change in the cluster namespace that holds cluster role and role bindings
-func (ac *AuthorizationCache) invalidateCache() bool {
-	invalidateCache := false
+func (ac *AuthorizationCache) invalidateCache(lastCacheInvalidation time.Time) bool {
```
benluddy, line 378: Shouldn't the parameter be the current time? There's no upside to passing a struct field value to a method on the struct itself, but either passing the current time or using a clock object to get the time prevents the tests from being dependent on the passage of real time (via calls to `time.Now` or `time.Since`).

benluddy: I notice you removed the deferred function, does that mean that it would perform a full invalidation back-to-back if the cluster RBAC changes at the same time as the TTL is reached?

pkg/project/auth/cache.go
```diff
@@ -395,11 +413,18 @@ func (ac *AuthorizationCache) invalidateCache() bool {
...
+// cacheHasExpired is used to evaluate when it is time to do a full cache
+// invalidation due to age.
+func (ac *AuthorizationCache) cacheHasExpired(lastCacheInvalidation time.Time) bool {
+	return time.Since(lastCacheInvalidation) > ac.maxCacheLifespan
+}
+
 // synchronize runs a a full synchronization over the cache data.  it must be run in a single-writer model, it's not thread-safe by design.
 func (ac *AuthorizationCache) synchronize() {
-	// if none of our internal reflectors changed, then we can skip reviewing the cache
+	// if none of our internal reflectors changed, and the cache hasn't
```
benluddy, line 424: If we need to look at the TTL to make decisions about both "skip" and "invalidate", would it be better to check the TTL exactly once at the beginning of `synchronize`, then pass an "expired" parameter to `invalidateCache`? That would make `invalidateCache` much simpler -- basically just initializing the bool with the argument -- and I think it'd be easier to understand because all the interesting logic is localized near the top of `synchronize`.

pkg/project/auth/cache_test.go
```diff
@@ -250,12 +251,24 @@ func TestInvalidateCache(t *testing.T) {
...
-			ac := &AuthorizationCache{
-				clusterRoleLister:        syncedClusterRoleLister{ClusterRoleLister: rbacv1listers.NewClusterRoleLister(crs)},
-				clusterRoleBindingLister: syncedClusterRoleBindingLister{ClusterRoleBindingLister: rbacv1listers.NewClusterRoleBindingLister(crbs)},
+			ac.clusterRoleLister = syncedClusterRoleLister{
+				ClusterRoleLister: rbacv1listers.NewClusterRoleLister(crs),
+				versioner:         informers.Rbac().V1().ClusterRoles().Informer(),
+			}
+			ac.clusterRoleBindingLister = syncedClusterRoleBindingLister{
+				ClusterRoleBindingLister: rbacv1listers.NewClusterRoleBindingLister(crbs),
+				versioner:                informers.Rbac().V1().ClusterRoleBindings().Informer(),
```
benluddy, line 271: Did all of this change in order to call `NewAuthorizationCache`? Is it necessary?

pkg/project/auth/cache_test.go
```diff
@@ -269,7 +282,7 @@ func TestInvalidateCache(t *testing.T) {
 						defer crbs.Delete(&trial.crbs[i])
 					}
 
-					actual := ac.invalidateCache()
+					actual := ac.invalidateCache(time.Now())
```
benluddy, line 285: The suggestion to pass a bool here instead would also mean simpler test setup.

pkg/project/auth/cache_test.go
```diff
@@ -279,3 +292,46 @@ func TestInvalidateCache(t *testing.T) {
...
+	reviewer := &mockReviewer{}
+	nsInformer := informers.Core().V1().Namespaces().Informer()
+	rbacClient := informers.Rbac().V1()
+	ac := NewAuthorizationCache(nsLister, nsInformer, reviewer, rbacClient)
+
+	if invalidate := ac.invalidateCache(time.Now()); invalidate {
+		t.Errorf("expected false on check first check, got true")
+	}
+
+	past := time.Now().Add(-2 * defaultMaxCacheLifespan)
```
benluddy, line 312: Won't this flake if more than 2 real seconds elapse between the first call to `time.Now` on L308 and the second on L312? It would be ideal to avoid calling `time.Now` or `time.Since` anywhere in this patch.

ricardomaraschini:
I don't think I can keep moving this one forward. Closing it.

pkg/project/auth/cache_test.go
ricardomaraschini: Yeah, again, not ideal. I was trying to make in such a way I could avoid using `time.Now()` in the tests. Let me see what I can do about this.

pkg/project/auth/cache.go
ricardomaraschini: Yes, it was not intentional. Every time I revisit this code I lost some context. Let me re-think about this.

pkg/project/auth/cache_test.go
ricardomaraschini: One would not expect it to happen but yeah, I guess there is a remote possibility. Back to the drawing board.

pkg/project/auth/cache_test.go
ricardomaraschini: Possible, yeah. The question is the function responsibility here I guess. I got betrayed by my understanding so let me rethink.

pkg/project/auth/cache.go
ricardomaraschini: My initial understanding of this function was correct but then I don't know why I moved out of that `defer()` approach. I think my head thought it was just returning "true" or "false" and not mangling with the object internal state. Need to rethink.

PR #547, opened 2025-09-01
OCPBUGS-57474: ensure cache invalidation after a time

ricardomaraschini:
this commit ensures that we invalidate the cache at least every 15 seconds. this has been the chosen workaround for an existing bug that is on our code since long ago but just recently bubbled up in our e2e tests.

ideally we would refactor the whole cache layer but giving the time constraints we have opted-in for a more conservative approach.

## Summary by CodeRabbit

* **New Features**
  * Authorization cache now expires and refreshes on a predictable, clock-driven schedule to reduce stale permissions.
  * Refresh logic skips work when the cache is still fresh, improving performance.

* **Tests**
  * Expanded deterministic tests using a mockable clock to validate time-driven expiration, refresh behavior, and edge cases.

### DIFF
--- a/pkg/project/auth/cache.go
+++ b/pkg/project/auth/cache.go
@@ -21,10 +21,16 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	rbacv1listers "k8s.io/client-go/listers/rbac/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/utils/clock"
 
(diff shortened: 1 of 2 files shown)


benluddy:
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.20-upgrade-from-stable-4.19-e2e-metal-ipi-ovn-upgrade 5

PR #550, opened 2025-09-04
OCPBUGS-61228: Bump Route generation when spec is updated

rikatz:
Kubernetes resources should bump their generation every time a field on spec is changed. This allows controllers to keep track of the desired state and the current state.

This change:
* initializes the generation of a resource if it does not exist
* updates the generation of a resource when a field on spec is changed
* adds some unit tests to assure the desired behavior


alebedev87:
> Maybe you and Miciah should also be added to approvers? :)

https://github.com/openshift/openshift-apiserver/pull/558

PR #558, opened 2025-09-17
NO-JIRA: route: Update OWNERS

alebedev87:
- OpenShift Ingress&DNS team added as reviewers and approvers.
- `smarterclayton`, `pweil`, `rajatchopra` and `ironcladlou` moved to emeritus approvers.
- `imcsk8`,`pravisankar` and `pecameron` removed from reviewers because they left Red Hat.
- `jacobtanenbaum` removed from reviewers (not OpenShift Ingress&DNS team).

### DIFF
--- a/pkg/route/OWNERS
+++ b/pkg/route/OWNERS
@@ -1,16 +1,29 @@
 reviewers:
-  - smarterclayton
-  - pweil-
-  - imcsk8
   - knobunc
-  - ironcladlou
-  - pecameron
-  - pravisankar
-  - rajatchopra
-  - jacobtanenbaum
+  - Miciah
+  - alebedev87
+  - rikatz
+  - candita
+  - rfredette
+  - gcs278
+  - davidesalerno
+  - bentito
+  - Thealisyed
+  - grzpiotrowski
 approvers:
+  - knobunc
+  - Miciah
+  - alebedev87
+  - rikatz
+  - candita
+  - rfredette
+  - gcs278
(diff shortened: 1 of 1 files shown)


alebedev87:
/assign @p0lyn0mial 
/assign @ingvagabund 

Folks, can you please have a look at this one? I'm missing an approval from the repository owners.

pkg/route/OWNERS
```diff
@@ -1,16 +1,29 @@
...
+  - rikatz
+  - candita
+  - rfredette
+  - gcs278
+  - davidesalerno
+  - bentito
+  - Thealisyed
+  - grzpiotrowski
+emeritus_approvers:
   - smarterclayton
```
ingvagabund, line 26: smarterclayton left few years back too

pkg/route/OWNERS
```diff
@@ -1,16 +1,29 @@
...
-  - smarterclayton
-  - pweil-
-  - imcsk8
   - knobunc
-  - ironcladlou
-  - pecameron
-  - pravisankar
-  - rajatchopra
-  - jacobtanenbaum
+  - Miciah
```
ingvagabund, line 3: There's a lot of names where I am not sure to which team each belongs to. What about to create an alias under https://github.com/openshift/openshift-apiserver/blob/main/OWNERS_ALIASES? Assuming everyone added into both pools are from the same network oriented team(s)?

PR #576, opened 2025-11-20
[release-4.19] OCPBUGS-61662: Adding validation directly and calling existing library-go validation

Thealisyed:
Add route path validation to reject # or whitespace
Add ValidatePath function and integrate it with ValidateRoute
and ValidateRouteUpdate to reject routes with '#' or spaces in spec.path

Validation changes:
- Add ValidatePath to check for # and spaces in route paths
- Call the library-go validation and add additional path validation

This ensures our PR matches the validation logic introduced
in https://github.com/openshift/library-go/pull/2012/files

### DIFF
--- a/pkg/route/apis/route/validation/validation.go
+++ b/pkg/route/apis/route/validation/validation.go
@@ -2,6 +2,7 @@ package validation
 
 import (
 	"context"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/route/apis/route/validation/validation.go
```diff
@@ -42,7 +58,14 @@ func ValidateRouteUpdate(ctx context.Context, route *routeapi.Route, oldRoute *r
 		return errs
 	}
 
-	return routevalidation.ValidateRouteUpdate(ctx, external, oldExternal, sarClient, secretsGetter, opts)
+	allErrs := routevalidation.ValidateRouteUpdate(ctx, external, oldExternal, sarClient, secretsGetter, opts)
+
+	if external.Spec.Path != oldExternal.Spec.Path {
+		pathFieldPath := field.NewPath("spec", "path")
+		allErrs = append(allErrs, validatePath(external.Spec.Path, pathFieldPath)...)
+	}
```
benluddy: OK, the update validation is ratcheting, so it won't block updates to preexisting routes with this problem.

PR #591, opened 2026-01-06
OCPBUGS-72408: block image import from cluster IP addresses

ricardomaraschini:
This PR adds validation to block image imports from `localhost`, `link-local` addresses (169.254.0.0/16), and cluster service/pod CIDR ranges. The implementation fetches service and pod CIDRs from the cluster Network configuration, performs DNS resolution to validate hostnames, and fails closed if the network config is unavailable.

### DIFF
--- a/pkg/image/apiserver/registry/imagestreamimport/rest.go
+++ b/pkg/image/apiserver/registry/imagestreamimport/rest.go
@@ -3,7 +3,9 @@ package imagestreamimport
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
+	"net/netip"
 	"strings"
 	"time"
 
@@ -25,6 +27,7 @@ import (
 	kapi "k8s.io/kubernetes/pkg/apis/core"
 	kapihelper "k8s.io/kubernetes/pkg/apis/core/helper"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/image"
 	imagev1 "github.com/openshift/api/image/v1"
 	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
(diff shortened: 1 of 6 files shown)


flavianmissi (commented):
Left a question/possible nit, but LGTM either way.

/lgtm

pkg/image/apiserver/registry/imagestreamimport/transport.go
```diff
@@ -0,0 +1,51 @@
...
+
+// RestrictedTransport is an http.RoundTripper that restricts requests to
+// certain IP ranges. It is meant to avoid doing requests to addresses the
+// registry should not be allowed to reach.
+type RestrictedTransport struct {
+	wrapped           http.RoundTripper
+	extraBlockedCIDRs []netip.Prefix
+}
+
+// RoundTrip implements the http.RoundTripper interface. If the IP is blocked
```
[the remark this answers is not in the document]
ricardomaraschini: Yes, it makes way more sense to enforce it at compile time.

ricardomaraschini:
@ingvagabund This is ready for another look.

pkg/image/apiserver/registry/imagestreamimport/rest.go
```diff
@@ -195,6 +200,115 @@ func (r *REST) createImages(
...
+			return nil, fmt.Errorf("invalid registry url: %w", err)
+		}
+
+		// now we know that we only miss the port in the registry. we
+		// set a zeroed out port so SplitHostPort can succeed. XXX we
+		// use this approach because SplitHostPort takes care of
+		// properly parsing IPv6 addresses and strips out the brackets.
+		withPort := fmt.Sprintf("%s:0", imageConfig.Status.InternalRegistryHostname)
+		if hostname, _, err = net.SplitHostPort(withPort); err != nil {
+			return nil, fmt.Errorf("invalid registry url: %w", err)
```
sanchezl: This might be a confusing message with the `:0`. Either strip it out of the message or use 443, the actual default port.

ricardomaraschini: We only end here if the error is because of a missing port so this isn't ever expected to fail. The zero was to demonstrate intent: we do not care about the port.

That being said: setting this to `443` made the LLM reviewer here (code rabbit) ask me to use `0` instead.

PR #599, opened 2026-01-30
Revert "OCPBUGS-72408: block image import from cluster IP addresses"

ricardomaraschini:
After some internal discussions we have opted for reverting this. We will be discussing what is the best approach to solve this problem.


ricardomaraschini:
/verified bypass

I think as this is a revert we can bypass the verification.

PR #600, opened 2026-02-02
[release-4.15] OCPBUGS-74923: bump kubernetes-apiserver to pick up loopback cert expiration update

everettraven:


p0lyn0mial:
This PR also pulls in https://github.com/kubernetes/kubernetes/pull/118886 , which avoids starting informers or the config-consuming controller when --enable-priority-and-fairness=false.

It does not enable or disable the PnF feature, the operator is responsible for that.

See also https://github.com/openshift/oauth-apiserver/pull/173#issuecomment-3943543163

PR #609, opened 2026-02-23
OCPBUGS-68208: CVE-2025-65637 - Bump github.com/sirupsen/logrus from v1.9.0 to v1.9.3 [release-4.17]

rissh:
## Description
Bump `github.com/sirupsen/logrus` to v1.9.3 to fix CVE-2025-65637 (DoS in logrus). Direct dependency update only; library-go is unchanged to avoid k8s version drift (per approach on [release-4.17 PR #609](https://github.com/openshift/openshift-apiserver/pull/609)).

## Changes
- Bumped `github.com/sirupsen/logrus` from v1.9.0 to v1.9.3

## CVE
- **CVE-2025-65637** | High (CVSS 7.5) | Fixed in v1.9.3

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -121,7 +121,7 @@ require (
 	github.com/prometheus/common v0.44.0 // indirect
 	github.com/prometheus/procfs v0.10.1 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
-	github.com/sirupsen/logrus v1.9.0 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/soheilhy/cmux v0.1.5 // indirect
 	github.com/stoewer/go-strcase v1.2.0 // indirect
 	github.com/stretchr/testify v1.8.4 // indirect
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -121,7 +121,7 @@ require (
 	github.com/prometheus/common v0.44.0 // indirect
 	github.com/prometheus/procfs v0.10.1 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
-	github.com/sirupsen/logrus v1.9.0 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
```
p0lyn0mial, line 124: Ah, the issue is that in 4.17, library-go is on 1.30, whereas openshift-apiserver is on 1.29.
For this PR, we might be able to force just the logrus dependency update.

But in the future we may need to pull other changes made directly in library-go, and at that point a kube bump will be unavoidable.

Would like to ask @benluddy @bertinatto for an opinion.

rissh: Exactly—I noticed that mismatch as well. That is specifically why I paused after 4.17, as I anticipated this drift would only become more significant as we move down toward 4.12

Definitely interested to hear what others think about this, since I am already facing this issue for other components as well I move down.

bertinatto: >  For this PR, we might be able to force just the logrus dependency update.

+1 for this. Let's just bump logrus to address the CVE.

p0lyn0mial: @rissh so instead of pulling the library-go we could just update the `github.com/sirupsen/logrus` dep to ` v1.9.3`.

I think this could work.

rissh: thanks @bertinatto for the ack, I will move with direct logrus bump then for this PR and as I move down towards 4.12

rissh: > @rissh so instead of pulling the library-go we could just update the `github.com/sirupsen/logrus` dep to ` v1.9.3`.
> 
> I think this could work.

yep, will do the same.

p0lyn0mial:
the diff looks much better, thanks.

PR #619, opened 2026-03-11
NO-JIRA: Send bookmark events to be compatible with watchlist

ardaguclu:
When request contains SendInitialEvents: true option (WatchList feature gate can send it), we need to emit all the projects just like we do when resourceVersion: 0. Additionally, in order to be compatible with WatchList feature gate, watch handler must return special bookmark event to signal that all the previous projects are emitted. 

This PR makes changes to make project watch handle compatible with WatchList.

## Summary by CodeRabbit

* **New Features**
  * Optional initial-events bookmark for project watches to mark completion of initial event delivery.

* **Updates**
  * Bookmark notifications now bypass normal filters so clients reliably detect initial-events completion.
  * Watcher records the cluster resource version and emits a clear end-of-initial-events bookmark for improved synchronization and debugging.


ardaguclu:
Changes here do not fully support WatchList
/close

PR #622, opened 2026-03-16
OCPBUGS-78596: Fix image reference in TestImageStreamImportQuayIO [release-4.13] 

jubittajohn:
quay.io/coreos/etcd:latest is gone, use a different tag instead.
cherry-picked using claude


sanchezl:
The fix looks correct — `quay.io/coreos/etcd:latest` no longer exists, pinning to `v3.4.37` is the right approach.

/lgtm
/approve
/verified bypass

PR #629, opened 2026-04-13
CNTRLPLANE-3226: Sync with openshift apiserver 4.22 kubernetes 1.34.1

ingvagabund:
Evidence PR for https://github.com/openshift/kubernetes-apiserver/pull/85

## Summary by CodeRabbit

* **Chores**
  * Updated module dependencies to newer versions for improved compatibility
  * Aligned kubernetes-apiserver references to OpenShift platform version 4.22
  * Refined OpenAPI schema validation rules across multiple API versions and resources
  * Updated build script configuration for OpenAPI code generation processes


hack/update-generated-openapi.sh
```diff
@@ -42,7 +42,7 @@ echo "Generating origin openapi"
 ${GOPATH}/bin/openapi-gen \
   --output-file zz_generated.openapi.go \
   --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt \
-  --output-dir="${GOPATH}/src" \
+  --output-dir="${GOPATH}/src/${ORIGIN_PREFIX}/pkg/openapi" \
```
p0lyn0mial, line 45: why do we need this change ? why it worked before ?

pkg/openapi/openapi_test.go
```diff
@@ -51,6 +57,24 @@ func TestOpenAPIRoundtrip(t *testing.T) {
 			delete(roundTripped.Extensions, common.ExtensionV2Schema)
 			delete(value.Schema.Extensions, common.ExtensionV2Schema)
 
+			// openapi-gen emits empty Enum fields as []interface{}{} in specific schemas,
```
p0lyn0mial: why it this normalization needed now ?

hack/update-generated-openapi.sh
ingvagabund: I think at some point `--output-dir` semantics changed? Without this change the file is generated under `$GOPATH/src`.

pkg/openapi/openapi_test.go
ingvagabund: Thank you for asking. The reason is quite suprising:
- https://github.com/openshift/api/blob/release-4.21/config/v1/types_authentication.go#L725-L727: `TokenValidationRuleTypeRequiredClaim` constant is untyped. After changing it to `TokenValidationRuleTypeRequiredClaim TokenValidationRuleType = "RequiredClaim"` the enum is no longer empty.
- https://github.com/openshift/api/blob/release-4.21/config/v1/types_authentication.go#L675-L685 are defined as variables, not constants. After changing them to constants the enum is no longer empty.

ingvagabund: After https://github.com/openshift/api/pull/2814 is merged and [backported to 4.21](https://github.com/openshift/api/pull/2815) the unit test "fix" can be dropped.

p0lyn0mial: should we wait for 2814 before merging this PR then ?

hack/update-generated-openapi.sh
p0lyn0mial: ok, thanks.

pkg/openapi/openapi_test.go
ingvagabund: Without https://github.com/openshift/api/pull/2815 merged the generated openapi is not correct for those two types. I'd rather wait so we don't need to double backport.

p0lyn0mial: ok, sgtm. please ping me one 2815 merges. thank you for all your work Jan.

ingvagabund: Done. PTAL.

PR #642, opened 2026-05-07
OCPBUGS-84534: fix concurrent map race in project authorization cache

sanchezl:
## Summary

Fix a fatal `concurrent map iteration and map write` panic in `AuthorizationCache` that intermittently crashes openshift-apiserver pods. The race has existed for years and was never successfully fixed.

**Dual-path copy-on-write with atomic store swap:**

During full cache invalidation (every 15s or on RBAC changes), new stores are built privately by the writer goroutine — mutations happen in place with zero copy overhead. During incremental updates between invalidations, stores are shared with concurrent `List()` readers — `addSubjectsToNamespace` and `deleteNamespaceFromSubjects` use copy-on-write to create new `subjectRecord` objects with copied namespaces sets, so readers iterate immutable snapshots. All three stores are grouped behind `atomic.Pointer[authorizationCacheStores]` so the full-rebuild swap is a single atomic operation — readers always see a consistent view.

**No-op guards** skip the COW copy entirely when the namespace is already present/absent from the subject's set — this eliminates unnecessary allocations when subjects are re-processed with unchanged access (common in incremental updates, especially with duplicate subjects from upstream).

This avoids the lock contention that caused the previous mutex fix (PR #267) to be reverted (PR #326), while also avoiding the O(n²) allocation overhead of unconditional copy-on-write that was identified during code review.

## Root Cause

`List()` (called from HTTP request goroutines via `proxy.(*REST).List`) reads `subjectRecord.namespaces` (`sets.String` = `map[string]Empty`). Meanwhile, `synchronize()` (background goroutine) mutates the same maps in place:
- `addSubjectsToNamespace()`: `item.namespaces.Insert(namespace)`
- `deleteNamespaceFromSubjects()`: `delete(subjectRecord.namespaces, namespace)`

Go's runtime detects the concurrent map read+write and kills the process with a fatal panic. Additionally, three store pointers (`userSubjectRecordStore`, `groupSubjectRecordStore`, `reviewRecordStore`) were swapped non-atomically during full cache rebuilds — a concurrent `List()` could read stores from different generations, returning silently wrong results.

## Fix Strategy

Two races and one performance issue are fixed:

1. **Map race → dual-path COW:** When stores are shared with readers (incremental updates), create new `subjectRecord` with a copied `sets.String` instead of mutating in place. When stores are private to the writer (full invalidation rebuild), mutate in place for performance. The `copyOnWrite bool` parameter threads through the call chain to select the appropriate path.

2. **Non-atomic store swap → `atomic.Pointer`:** Group all three stores in `authorizationCacheStores` behind `atomic.Pointer`. Full rebuilds populate private stores, then swap atomically. `List()` snapshots the pointer once at entry.

3. **No-op guards:** In the COW path, check `Has(namespace)` / `!Has(namespace)` before copying. When a subject already has (or already lacks) access to a namespace, skip the copy entirely. This is critical because an [upstream kube bug](https://github.com/kubernetes/kubernetes/blob/04976d6f60b883911a2ee6df12a54864391bd574/plugin/pkg/auth/authorizer/rbac/subject_locator.go#L109-L124) in `AllowedSubjects()` returns duplicate subjects (line 124 returns `subjects` instead of `dedupedSubjects`), causing each duplicate to trigger a redundant COW copy.

## Alternatives Considered

Three approaches were prototyped and benchmarked (see [comment below](https://github.com/openshift/openshift-apiserver/pull/642#issuecomment-4434066208) for full comparison):

| Approach | 1000ns / 100u | Allocs | Branch |
|----------|--------------|--------|--------|
| **Dual-path COW** (this PR) | ~9ms / 15MB | 16K | [`bugfix/project-auth-cache-race`](https://github.com/sanchezl/openshift-apiserver/tree/bugfix/project-auth-cache-race) |
| sync.Map | ~12ms / 12MB | 27K | [`bugfix/project-auth-cache-race-syncmap`](https://github.com/sanchezl/openshift-apiserver/tree/bugfix/project-auth-cache-race-syncmap) |
| Fine-grained RWMutex | ~12ms / 11MB | 27K | [`bugfix/project-auth-cache-race-rwmutex`](https://github.com/sanchezl/openshift-apiserver/tree/bugfix/project-auth-cache-race-rwmutex) |

All three are viable. Dual-path COW was chosen because it is the fastest, fully lock-free on the read path, and carries no risk of reintroducing the lock contention that caused the PR #326 revert.

## Fix History

1. **PR #267** (Jan 2022): Added `sync.RWMutex` to synchronize access
2. **PR #326** (Oct 2022): **Reverted** the mutex — clusters with high namespace/RBAC counts had multi-minute sync times, blocking all `List()` requests (goroutine dumps showed 3-4 minute waits on `RLock`)
3. **PR #547** (Sep 2025): Timer-based cache invalidation every 15s for OCPBUGS-57474, but no locking — the race remained
4. **PR #530**: "Authorization Cache V2" full rewrite — abandoned
5. **This PR**: Dual-path COW + atomic pointer + no-op guards — `List()` never blocks, no O(n²) regression

## Related Issues

- **OCPBUGS-65926** — Same crash on 4.14.26, incorrectly closed as duplicate of OCPBUGS-56594 (a different bug in kube-apiserver audit log serialization)
- **OCPBUGS-58029** — Clone of 65926
- **OCPBUGS-56594** — Different bug (kube-apiserver audit race), fixed by kubernetes/kubernetes#129472
- **Upstream kube bug** — [`subject_locator.go:124`](https://github.com/kubernetes/kubernetes/blob/04976d6f60b883911a2ee6df12a54864391bd574/plugin/pkg/auth/authorizer/rbac/subject_locator.go#L124) returns `subjects` instead of `dedupedSubjects`, causing duplicate subjects to flow into `addSubjectsToNamespace`. Mitigated by no-op guards in this PR; upstream fix tracked separately.

### Test 1: Race condition is fixed
- Deploy a build with the fix
- Run the cluster under load with concurrent project list requests
- Verify no panics in openshift-apiserver logs and no pod restarts with exitCode 2

### Test 2: No regression on large clusters (the PR #267 revert scenario)
- Provision a cluster with **2000+ namespaces** and substantial RBAC
- Measure `oc get projects` latency before and after — should not regress
- Key property: `List()` should never block waiting on the sync goroutine
- Monitor openshift-apiserver memory usage for unexpected growth

### Test 3: Cache freshness
- Grant/revoke a user's access to a namespace
- Verify reflected in `oc projects` output within ~15 seconds

## Verification

`/verified by "TestAuthorizationCacheRace"`

## Summary by CodeRabbit

* **Refactor**
  * Redesigned the authorization cache to provide consistent snapshot reads and safe concurrent updates using atomic store replacement and copy-on-write semantics.

* **Tests**
  * Added a concurrency race test exercising simultaneous readers/writers against the cache.
  * Added benchmarks for full cache invalidation, incremental copy-on-write behavior, and subject-update performance across varying namespace counts.

### DIFF
--- a/pkg/project/auth/cache.go
+++ b/pkg/project/auth/cache.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -173,6 +174,14 @@ func (l syncedClusterRoleBindingLister) LastSyncResourceVersion() string {
 	return l.versioner.LastSyncResourceVersion()
 }
 
+// authorizationCacheStores groups the three cache stores so they can be
+// swapped atomically during full cache invalidation.
+type authorizationCacheStores struct {
+	reviewRecordStore       cache.Store
+	userSubjectRecordStore  cache.Store
+	groupSubjectRecordStore cache.Store
+}
+
 // AuthorizationCache maintains a cache on the set of namespaces a user or group can access.
 type AuthorizationCache struct {
 	// allKnownNamespaces we track all the known namespaces, so we can detect deletes.
@@ -187,9 +196,7 @@ type AuthorizationCache struct {
 	roleBindingNamespacer         SyncedRoleBindingLister
 	roleLastSyncResourceVersioner LastSyncResourceVersioner
 
-	reviewRecordStore       cache.Store
-	userSubjectRecordStore  cache.Store
-	groupSubjectRecordStore cache.Store
+	stores atomic.Pointer[authorizationCacheStores]
 
 	clusterBindingResourceVersions sets.String
 	clusterRoleResourceVersions    sets.String
@@ -199,7 +206,7 @@ type AuthorizationCache struct {
 
 	reviewer Reviewer
 
-	syncHandler func(request *reviewRequest, userSubjectRecordStore cache.Store, groupSubjectRecordStore cache.Store, reviewRecordStore cache.Store) error
+	syncHandler func(request *reviewRequest, userSubjectRecordStore cache.Store, groupSubjectRecordStore cache.Store, reviewRecordStore cache.Store, copyOnWrite bool) error
 
 	watchers    []CacheWatcher
 	watcherLock sync.Mutex
@@ -255,10 +262,6 @@ func NewAuthorizationCache(
 		roleBindingNamespacer:         srbLister,
 		roleLastSyncResourceVersioner: unionLastSyncResourceVersioner{scrLister, scrbLister, srLister, srbLister},
 
-		reviewRecordStore:       cache.NewStore(reviewRecordKeyFn),
-		userSubjectRecordStore:  cache.NewStore(subjectRecordKeyFn),
-		groupSubjectRecordStore: cache.NewStore(subjectRecordKeyFn),
-
 		reviewer: reviewer,
 		skip:     &neverSkipSynchronizer{},
 
@@ -268,6 +271,11 @@ func NewAuthorizationCache(
 		lastCacheInvalidation: realClock.Now(),
 		maxCacheLifespan:      defaultMaxCacheLifespan,
 	}
+	ac.stores.Store(&authorizationCacheStores{
+		reviewRecordStore:       cache.NewStore(reviewRecordKeyFn),
+		userSubjectRecordStore:  cache.NewStore(subjectRecordKeyFn),
+		groupSubjectRecordStore: cache.NewStore(subjectRecordKeyFn),
+	})
 	ac.lastSyncResourceVersioner = namespaceLastSyncResourceVersioner
 	ac.syncHandler = ac.syncRequest
 	return ac
@@ -315,7 +323,7 @@ func (ac *AuthorizationCache) GetClusterRoleLister() SyncedClusterRoleLister {
 }
 
 // synchronizeNamespaces synchronizes access over each namespace and returns a set of namespace names that were looked at in last sync
-func (ac *AuthorizationCache) synchronizeNamespaces(userSubjectRecordStore cache.Store, groupSubjectRecordStore cache.Store, reviewRecordStore cache.Store) sets.String {
+func (ac *AuthorizationCache) synchronizeNamespaces(userSubjectRecordStore cache.Store, groupSubjectRecordStore cache.Store, reviewRecordStore cache.Store, copyOnWrite bool) sets.String {
 	namespaceSet := sets.NewString()
 	namespaces, err := ac.namespaceLister.List(labels.Everything())
 	if err != nil {
@@ -329,7 +337,7 @@ func (ac *AuthorizationCache) synchronizeNamespaces(userSubjectRecordStore cache
 			namespace:                namespace.Name,
 			namespaceResourceVersion: namespace.ResourceVersion,
 		}
-		if err := ac.syncHandler(reviewRequest, userSubjectRecordStore, groupSubjectRecordStore, reviewRecordStore); err != nil {
+		if err := ac.syncHandler(reviewRequest, userSubjectRecordStore, groupSubjectRecordStore, reviewRecordStore, copyOnWrite); err != nil {
 			utilruntime.HandleError(fmt.Errorf("error synchronizing: %v", err))
 		}
 	}
@@ -337,7 +345,7 @@ func (ac *AuthorizationCache) synchronizeNamespaces(userSubjectRecordStore cache
 }
 
 // synchronizePolicies synchronizes access over each role
-func (ac *AuthorizationCache) synchronizePolicies(userSubjectRecordStore cache.Store, groupSubjectRecordStore cache.Store, reviewRecordStore cache.Store) {
+func (ac *AuthorizationCache) synchronizePolicies(userSubjectRecordStore cache.Store, groupSubjectRecordStore cache.Store, reviewRecordStore cache.Store, copyOnWrite bool) {
 	roleList, err := ac.roleNamespacer.Roles(metav1.NamespaceAll).List(labels.Everything())
 	if err != nil {
 		utilruntime.HandleError(err)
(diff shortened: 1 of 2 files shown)


pkg/project/auth/cache.go
```diff
@@ -600,10 +614,13 @@ func deleteNamespaceFromSubjects(subjectRecordStore cache.Store, subjects []stri
 	for _, subject := range subjects {
 		obj, exists, _ := subjectRecordStore.GetByKey(subject)
 		if exists {
-			subjectRecord := obj.(*subjectRecord)
-			delete(subjectRecord.namespaces, namespace)
-			if len(subjectRecord.namespaces) == 0 {
-				subjectRecordStore.Delete(subjectRecord)
+			old := obj.(*subjectRecord)
+			newNamespaces := sets.NewString(old.namespaces.UnsortedList()...)
```
benluddy: I'm concerned about this from a performance perspective -- especially allocations -- because this will run something like O(N*U) times on each invalidation (N=namespaces and U=users).

sanchezl:
## Alternative Approaches Evaluated

Three approaches were prototyped independently and benchmarked using `BenchmarkFullCacheInvalidation` (added by @benluddy in commit 2). Each approach fixes both the `concurrent map iteration and map write` panic and the non-atomic store pointer swap.

### 1. Dual-path COW (this PR)

**Branch:** [`bugfix/project-auth-cache-race`](https://github.com/sanchezl/openshift-apiserver/tree/bugfix/project-auth-cache-race)

During full cache invalidation, new stores are private to the writer — mutations happen in place with zero copy overhead. During incremental updates, stores are shared with readers — copy-on-write creates new `subjectRecord` objects with copied `sets.String`. A `copyOnWrite bool` parameter threads through the call chain. All three stores are grouped behind `atomic.Pointer[authorizationCacheStores]` for atomic swap.

### 2. sync.Map

**Branch:** [`bugfix/project-auth-cache-race-syncmap`](https://github.com/sanchezl/openshift-apiserver/tree/bugfix/project-auth-cache-race-syncmap)

Replace `sets.String` (backed by `map[string]sets.Empty`) with `sync.Map` for the `subjectRecord.namespaces` field. `sync.Map` provides built-in concurrent read/write safety without external locking. `List()` uses `Range()` to iterate, `addSubjectsToNamespace` uses `Store()`, `deleteNamespaceFromSubjects` uses `Delete()`.

### 3. Fine-grained RWMutex

**Branch:** [`bugfix/project-auth-cache-race-rwmutex`](https://github.com/sanchezl/openshift-apiserver/tree/bugfix/project-auth-cache-race-rwmutex)

Add a per-record `sync.RWMutex` to `subjectRecord`. Lock only individual map mutations (nanosecond-scale), not the entire `synchronize()` call (which can take minutes on large clusters). This is fundamentally different from the reverted PR #267, which held a single coarse-grained write lock across all of `synchronize()`.

### Benchmark Results

`BenchmarkFullCacheInvalidation` at 1000 namespaces × 100 users (the critical scale point):

| Approach | Time | Memory | Allocs |
|----------|------|--------|--------|
| **Dual-path COW** | **~9 ms** | **~15 MB** | **~16K** |
| sync.Map | ~12 ms | ~11 MB | ~27K |
| Fine-grained RWMutex | ~12 ms | ~11 MB | ~27K |
| Original COW-only (before optimization) | ~941 ms | ~3 GB | ~2M |

All three optimized approaches are within ~30% of each other and are orders of magnitude better than the original unconditional COW approach that @benluddy identified as O(N×U) catastrophic.

### Why Dual-path COW?

- **Fastest** at the critical scale point (~9ms vs ~12ms)
- **Fully lock-free on the read path** — `List()` never blocks, never contends. This property eliminates any risk of reintroducing the lock contention that caused PR #326's revert (where `RLock` waits reached 3-4 minutes)
- **No new concurrency primitives** in hot paths — the `sync.Map` approach introduces `Range()` iteration overhead per `List()` call; the RWMutex approach introduces lock/unlock per record mutation
- **Atomic pointer swap** ensures readers see a consistent snapshot of all three stores

The sync.Map and fine-grained RWMutex approaches are both viable alternatives if the team prefers a different tradeoff. The branches are available for review.

sanchezl:
@coderabbitai resume

pkg/project/auth/cache.go
sanchezl: See https://github.com/openshift/openshift-apiserver/pull/642#issuecomment-4434066208

sanchezl:
## No-op Guard Optimization + Upstream Kube Bug

Following @benluddy's review, this push adds two improvements:

### 1. No-op guards in COW path

`addSubjectsToNamespace` and `deleteNamespaceFromSubjects` now check `Has(namespace)` / `!Has(namespace)` before copying. When a subject's access hasn't changed (the common case in incremental updates), the COW copy is skipped entirely.

### 2. Upstream kube bug identified

[`subject_locator.go:124`](https://github.com/kubernetes/kubernetes/blob/04976d6f60b883911a2ee6df12a54864391bd574/plugin/pkg/auth/authorizer/rbac/subject_locator.go#L109-L124) builds a `dedupedSubjects` slice but returns the original `subjects` — discarding the dedup work. Duplicate subjects flow through `RBACSubjectsToUsersAndGroups` (no dedup) into `addSubjectsToNamespace`, causing redundant COW copies. The no-op guards mitigate this; an upstream fix is tracked separately.

### Benchmark: `BenchmarkIncrementalSyncDuplicateSubjects`

Exercises the incremental COW path with duplicate subjects (`D=1` = no dupes, `D=10` = 10x dupes per user):

**Before (no-op guards):**

| Scale | Time | Memory | Allocs |
|-------|------|--------|--------|
| N=1000, U=100, D=1 | 2.05s | 7.1 GB | 824K |
| N=1000, U=100, D=10 | 20.5s | 71.1 GB | 8.0M |

**After (with no-op guards):**

| Scale | Time | Memory | Allocs |
|-------|------|--------|--------|
| N=1000, U=100, D=1 | **130ms** | **209 KB** | **1K** |
| N=1000, U=100, D=10 | **135ms** | **215 KB** | **1K** |

The no-op guards eliminate ~99.99% of allocations in the incremental path. Duplicates no longer matter because the guard catches the no-op before any copying occurs.

sanchezl:
/cherrypick release-4.22 release-4.21 release-4.20 release-4.19 release-4.18

PR #656, opened 2026-06-08
OCPBUGS-79070: Fix: Preserve annotations when creating ImageStreamTag

ShazaAldawamneh:
Problem
Annotations in metadata.annotations were lost when creating an ImageStreamTag declaratively. The second apply would work, but not the first.

Root Cause
The Create method was missing the annotation assignment that the Update method had (line 357).

Solution
Added tagRef.Annotations = istag.Annotations in the Create method to match Update behavior.

Testing
Added test: TestCreateImageStreamTagPreservesAnnotations

## Summary by CodeRabbit

* **Bug Fixes**
  * Tag-level annotations are now populated from top-level metadata during image stream tag creation when appropriate, preserving compatibility.

* **Tests**
  * Added tests covering creation with annotations and tightened assertions to compare full created objects (handling dynamic fields).
  * Updated tests to use proper test contexts for more reliable create/update/retry behavior.

### DIFF
--- a/pkg/image/apiserver/registry/imagestreamtag/rest_test.go
+++ b/pkg/image/apiserver/registry/imagestreamtag/rest_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	etcd "go.etcd.io/etcd/client/v3"
 	authorizationapi "k8s.io/api/authorization/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -533,11 +535,18 @@ func TestCreateImageStreamTag(t *testing.T) {
 					Name:        "test:tag",
 					Annotations: map[string]string{},
 				},
-				Tag: &imageapi.TagReference{
-					Name:            "tag",
-					From:            &kapi.ObjectReference{Kind: "DockerImage", Name: "foo/bar/baz"},
-					ReferencePolicy: imageapi.TagReferencePolicy{Type: imageapi.SourceTagReferencePolicy},
-				},
+				Tag: func() *imageapi.TagReference {
+					gen := int64(1)
+					return &imageapi.TagReference{
+						Name:       "tag",
+						From:       &kapi.ObjectReference{Kind: "DockerImage", Name: "foo/bar/baz"},
+						Generation: &gen,
+						ImportPolicy: imageapi.TagImportPolicy{
+							ImportMode: imageapi.ImportModeLegacy,
+						},
+						ReferencePolicy: imageapi.TagReferencePolicy{Type: imageapi.SourceTagReferencePolicy},
+					}
+				}(),
 			},
 		},
 		"invalid tag": {
@@ -589,11 +598,22 @@ func TestCreateImageStreamTag(t *testing.T) {
 						"description": "test tag",
 					},
 				},
-				Tag: &imageapi.TagReference{
-					Name:            "tag",
-					From:            &kapi.ObjectReference{Kind: "DockerImage", Name: "foo/bar/baz"},
-					ReferencePolicy: imageapi.TagReferencePolicy{Type: imageapi.SourceTagReferencePolicy},
-				},
+				Tag: func() *imageapi.TagReference {
+					gen := int64(1)
+					return &imageapi.TagReference{
+						Name: "tag",
+						Annotations: map[string]string{
+							"set":         "true",
+							"description": "test tag",
+						},
+						From:       &kapi.ObjectReference{Kind: "DockerImage", Name: "foo/bar/baz"},
+						Generation: &gen,
+						ImportPolicy: imageapi.TagImportPolicy{
+							ImportMode: imageapi.ImportModeLegacy,
+						},
+						ReferencePolicy: imageapi.TagReferencePolicy{Type: imageapi.SourceTagReferencePolicy},
+					}
+				}(),
 			},
 		},
 	}
@@ -636,21 +656,13 @@ func TestCreateImageStreamTag(t *testing.T) {
 				// Verify expected output for all successful creates
 				if tc.expectedOutput != nil {
 					createdISTag := result.(*imageapi.ImageStreamTag)
-					expected := tc.expectedOutput.DeepCopy()
 
-					// Copy dynamic fields from actual result
-					expected.ResourceVersion = createdISTag.ResourceVersion
-					expected.UID = createdISTag.UID
-					expected.CreationTimestamp = createdISTag.CreationTimestamp
-					expected.Generation = createdISTag.Generation
-					expected.Labels = createdISTag.Labels
-					expected.Image = createdISTag.Image
-					expected.LookupPolicy = createdISTag.LookupPolicy
-					expected.Conditions = createdISTag.Conditions
-					expected.Tag = createdISTag.Tag
-
-					if !reflect.DeepEqual(expected, createdISTag) {
-						t.Errorf("%s: output mismatch\nexpected: %#v\ngot: %#v", name, expected, createdISTag)
+					// Use cmp.Diff to compare, ignoring only the truly dynamic fields
+					diff := cmp.Diff(tc.expectedOutput, createdISTag,
+						cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion", "UID", "CreationTimestamp", "Generation"),
+					)
+					if diff != "" {
+						t.Errorf("%s: output mismatch (-expected +actual):\n%s", name, diff)
 					}
 				}
 			}
--- a/pkg/image/apiserver/registry/imagestreamtag/rest_test.go
+++ b/pkg/image/apiserver/registry/imagestreamtag/rest_test.go
@@ -512,6 +512,7 @@ func TestCreateImageStreamTag(t *testing.T) {
 		expectError     bool
 		errorTargetKind string
 		errorTargetID   string
+		expectedOutput  *imageapi.ImageStreamTag
 	}{
 		"valid istag": {
 			istag: &imageapi.ImageStreamTag{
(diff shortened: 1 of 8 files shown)


ehearne-redhat:
/lgtm 

Great job on this! 🥳

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -590,6 +590,79 @@ func TestCreateImageStreamTag(t *testing.T) {
 	}
 }
 
+func TestCreateImageStreamTagPreservesAnnotations(t *testing.T) {
+	client, server, storage := setup(t, nil)
+	defer server.Terminate(t)
+
+	// Create the base imagestream
+	client.Put(
+		context.TODO(),
```
rh-roman: You could probably use the text context here

```suggestion
		t.Context(),
```

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -196,7 +196,10 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 			return nil, kapierrors.NewAlreadyExists(imagegroup.Resource("imagestreamtag"), istag.Name)
 		}
 		if istag.Tag != nil {
-			target.Spec.Tags[imageTag] = *istag.Tag
+			tagRef := *istag.Tag
+			tagRef.Name = imageTag
```
everettraven: Looking at https://github.com/openshift/openshift-apiserver/blob/22f07f6e7825c572abf8214e6a376078d9048296/pkg/image/apiserver/registry/imagestreamtag/rest_test.go#L516-L529 , it looks like creating an imagestreamtag with a tag reference with a different tag name should be allowed.

As-is, I think that this would result in always overwriting any non-matching but explicitly defined tag references to be the tag from the metadata.name field. I think this would result in potentially incorrect tag mapping behavior. Unfortunately, our existing test cases don't seem to actually test that the output of the create action are what we expect.

pkg/image/apiserver/registry/imagestreamtag/rest.go
```diff
@@ -196,7 +196,10 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation
...
+			tagRef.Annotations = istag.Annotations
```
everettraven: Instead of making the change here, I wonder if this change is better suited to being made in https://github.com/openshift/openshift-apiserver/blob/22f07f6e7825c572abf8214e6a376078d9048296/pkg/image/apiserver/registry/imagestreamtag/strategy.go#L40-L48
since this is more akin to defaulting the fields?

If we made the change in the `PrepareForCreate` method of the strategy to set the annotations on the tag reference only if they were unset and annotations are set on the imagestreamtag, the pre-existing logic would then be correct.

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -590,6 +590,79 @@ func TestCreateImageStreamTag(t *testing.T) {
 	}
 }
 
+func TestCreateImageStreamTagPreservesAnnotations(t *testing.T) {
```
everettraven: Instead of creating an entirely new test function, should we just add new entries to the existing `TestCreateImageStreamTag` test function above this one?

pkg/image/apiserver/registry/imagestreamtag/rest.go
ShazaAldawamneh: I added `tagRef.Name = imageTag` because I saw the Update method does the same thing at line 358. I thought I should match that behavior for consistency.

However, looking at the existing test you referenced, it seems like Create might have previously allowed Tag.Name to differ from the metadata tag name. If that's the case, adding this line would change existing behavior.

we have two options:
1. should we keep `tagRef.Name = imageTag` to match Update's behavior (forces tag names to match)?
2. or remove it and only fix the annotation issue (preserves any existing behavior where Tag.Name can differ)

Which approach do you prefer? I'm happy to remove the `tagRef.Name = imageTag` line and focus purely on the annotation fix if that's safer.

pkg/image/apiserver/registry/imagestreamtag/rest.go
ShazaAldawamneh: that make sense actually. thanks for pointing this out.

pkg/image/apiserver/registry/imagestreamtag/rest.go
everettraven: It looks like conversion behaviors for an ImageStream already silently does this based on https://github.com/openshift/openshift-apiserver/blob/22f07f6e7825c572abf8214e6a376078d9048296/pkg/image/apis/image/v1/conversion.go#L326

From what I can tell, it does look like the original intent is that the tag reference name must match that of the tag parsed from the `metadata.name` field of the ImageStreamTag.

With that in mind, this should be fine as-is.

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -585,7 +603,20 @@ func TestCreateImageStreamTag(t *testing.T) {
 					t.Errorf("%s: unexpected status: %#v", name, status.Details)
 					return
 				}
+			} else {
+				// Verify annotations are preserved for the test case that has them
+				createdISTag := result.(*imageapi.ImageStreamTag)
+				if name == "create with annotations" {
```
everettraven: Instead of checking for a specific test name to compare the output for, I think it would be better to update all the existing creation tests to also have test expectations as to the output of the create call.

Then we can have common logic that applies to each test case within this test.

ShazaAldawamneh: Good suggestion! I've updated the test structure to include expectedAnnotations for all test cases and replaced the name checking logic with common verification that applies to all tests. This makes it easier to add new test cases and verify their output consistently.

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -508,10 +508,11 @@ func TestDeleteImageStreamTag(t *testing.T) {
...
 	tests := map[string]struct {
-		istag           runtime.Object
-		expectError     bool
-		errorTargetKind string
-		errorTargetID   string
+		istag               runtime.Object
+		expectError         bool
+		errorTargetKind     string
+		errorTargetID       string
+		expectedAnnotations map[string]string
```
everettraven: I think for the expected output, we should use a full `ImageStreamTag`. This makes it so that we have the foundation in place to compare the entire expected output for each of these test cases.

This helps to better set the expectations / intent that we can pull from these test cases when reading them in the future.

ShazaAldawamneh: I've updated the test to validate against a complete `ImageStreamTag` object rather than only checking annotations. The test now:

1. Creates a deep copy of the expected output
2. Copies dynamic fields from the actual result (such as `ResourceVersion`, `UID`, `CreationTimestamp`, etc.) into the expected object
3. Compares the full `ImageStreamTag` objects to ensure the entire resource matches the expected state

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -526,6 +527,18 @@ func TestCreateImageStreamTag(t *testing.T) {
...
 				},
 			},
+			expectedOutput: &imageapi.ImageStreamTag{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace:   "default",
+					Name:        "test:tag",
+					Annotations: map[string]string{},
+				},
+				Tag: &imageapi.TagReference{
+					Name:            "latest",
```
everettraven: Maybe I misunderstood from previous reviews, but I thought the expectation here is that this should always end up being the same as the tag parsed from the `metadata.name` of the `ImageStreamTag`. Is this not the case?

ShazaAldawamneh: You're absolutely right! The Tag.Name should be normalized to match the tag parsed from metadata.name.

I missed the update for this test case I fixed it now!

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -585,7 +632,29 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+					// Copy dynamic fields from actual result
+					expected.ResourceVersion = createdISTag.ResourceVersion
+					expected.UID = createdISTag.UID
+					expected.CreationTimestamp = createdISTag.CreationTimestamp
+					expected.Generation = createdISTag.Generation
+					expected.Labels = createdISTag.Labels
+					expected.Image = createdISTag.Image
+					expected.LookupPolicy = createdISTag.LookupPolicy
+					expected.Conditions = createdISTag.Conditions
+					expected.Tag = createdISTag.Tag
```
everettraven: This looks like it pretty much just sets all the things we would want to verify are in the expected output to match what the actual output is. This seems like it defeats the purpose of having the comparison.

I think using something like `cmp.Diff(expected, actual)` along with https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields to ignore _specific_ fields (like all the generated metadata) would be more accurate here.

ShazaAldawamneh: I wasn't aware of cmpopts.IgnoreFields, thanks for introducing me to it. Updated to use cmp.Diff with cmpopts.IgnoreFields to only ignore the dynamic metadata fields while properly verifying everything else. Ran go mod vendor to add cmpopts.

pkg/image/apiserver/registry/imagestreamtag/rest_test.go
```diff
@@ -548,6 +570,51 @@ func TestCreateImageStreamTag(t *testing.T) {
...
+				Tag: func() *imageapi.TagReference {
+					gen := int64(1)
+					return &imageapi.TagReference{
+						Name: "tag",
+						Annotations: map[string]string{
+							"set":         "true",
+							"description": "test tag",
+						},
+						From:       &kapi.ObjectReference{Kind: "DockerImage", Name: "foo/bar/baz"},
+						Generation: &gen,
```
everettraven, line 610: Nit: if you made this an IIFE because you needed to set this to a pointer to an int64, you could also just use `ptr.To(int64(1))` here and avoid the IIFE altogether.

everettraven (approved):
A very minor nit, but I won't block on it.

LGTM.

/lgtm

